PR #3, opened 2017-02-06
Computer vision profile test image configuration + better handling of web cameras to the main profile

ipuustin:
Add test configuration to computer vision profile. If the image is build in "development" mode, add interactive test and development packages. The initial test content is bit limited, but more is planned.

Also add libva-intel-driver and gstreamer-plugins-good to the computer vision images in both production and development mode. This enhances webcam support (by adding v4l2src element among other things) and enables HW acceleration of libva operations using Intel GPU backend.

### DIFF
--- a/meta-refkit/recipes-image/images/refkit-image-computervision.bb
+++ b/meta-refkit/recipes-image/images/refkit-image-computervision.bb
@@ -6,7 +6,10 @@ REFKIT_IMAGE_COMPUTERVISION_EXTRA_INSTALL ?= "${REFKIT_IMAGE_INSTALL_COMMON}"
 REFKIT_IMAGE_EXTRA_FEATURES += "${REFKIT_IMAGE_COMPUTERVISION_EXTRA_FEATURES}"
 REFKIT_IMAGE_EXTRA_INSTALL += "${REFKIT_IMAGE_COMPUTERVISION_EXTRA_INSTALL}"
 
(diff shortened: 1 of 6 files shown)


meta-refkit/recipes-image/images/refkit-image-computervision.bb
```diff
@@ -6,7 +6,13 @@ REFKIT_IMAGE_COMPUTERVISION_EXTRA_INSTALL ?= "${REFKIT_IMAGE_INSTALL_COMMON}"
 REFKIT_IMAGE_EXTRA_FEATURES += "${REFKIT_IMAGE_COMPUTERVISION_EXTRA_FEATURES}"
 REFKIT_IMAGE_EXTRA_INSTALL += "${REFKIT_IMAGE_COMPUTERVISION_EXTRA_INSTALL}"
 
-REFKIT_IMAGE_COMPUTERVISION_EXTRA_INSTALL_append = " opencv gstreamer1.0-vaapi"
+FEATURE_PACKAGES_computervision += "packagegroup-computervision"
```
mythi: how about adding these in refkit-image.bb?

ipuustin: The feature definition you mean? I can do that too if you think that's better. It's a tradeoff between encapsulating the computer vision stuff and making it easier to use several features in one image.

mythi: Yes, ```FEATURE_PACKAGES```. It's better to keep everything in the same place

ipuustin:
Pushed a new version of the patch set.

PR #5, opened 2017-02-07
[Do Not Merge] Include nodejs-runtime-tools in the gateway profile

nagineni:
Include nodejs-runtime-tools in the gateway profile.


mythi:
@nagineni is this still needed?

nagineni:
We are done with this test-build. Closing.

PR #6, opened 2017-02-07
Enable RMC

mythi:
This PR enables Runtime Machine Configuration (RMC) feature from meta-intel. It helps to apply board specific settings (e.g., kernel cmdline) runtime. In particular, we get the serial console information for the boards we're running.

### DIFF
--- a/meta-refkit/classes/image-dsk.bbclass
+++ b/meta-refkit/classes/image-dsk.bbclass
@@ -33,6 +33,7 @@ IMAGE_NAME_SUFFIX = ""
 
 do_uefiapp[depends] += " \
                          systemd-boot:do_deploy \
+                         rmc-db:do_deploy \
                          virtual/kernel:do_deploy \
                          initramfs-framework:do_populate_sysroot \
                          intel-microcode:do_deploy \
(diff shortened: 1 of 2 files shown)


mythi:
@jlaako @ipuustin FYI

pohly:
How does this feature interact with Secure Boot? If the signed UEFI app reads and applies the RMC database without checking it, then an attacker can easily gain control over the device by modifying the database.

I see two ways for avoiding this:
- validate the database
- disable the feature in images which are meant to be signed, ideally without having to change distro 
  settings

mythi:
@pohly I was aware of that open but no final answer how to address it. Does the PR look OK otherwise?

pohly:
mythi: I'm a bit uncomfortable with adding something that is enabled by default and that will break Secure Boot. Someone taking refkit for a product might forget to disable it.

Please file a bug about this so that we remember that something needs to be done.

Then ideally add a TODO item at the appropriate place (refkit.conf?) with a link to that bug before merging, if it's not too much work.

But I'm okay with merging it either way, as it seems to be urgent.

mythi:
@pohly thanks, it's not that urgent so I'll add a note about that.

mythi:
@pohly I've now added a note about the problem and filed a bugzilla entry to change the patch a little

PR #8, opened 2017-02-08
Make profile specific test manifests

skuusela:
Generate a test manifest for images by adding common tests and profile
specific tests to a manifest.


skuusela:
@mythi Feedback? Before merging this I will have to change the way DAFT executes tests because atm it's hardcoded on using "minnowmax.iottest.manifest".

skuusela:
test this please

mythi:
@skuusela this looks simple enough and OK to start with. 

About the hard-coded .manifest. Now that the test-suite becomes image specific, could you simplify it and come up with some common name for a test queue manifest that DAFT uses (i.e., say ```manifest_name``` is, e.g., testplan.manifest)

Another thing we need to keep in mind but not solve right now is how Testopia comes in the picture.

skuusela:
@mythi Rebased, made the script a bit prettier and changed the `manifest_name` to "image-testplan.manifest". I thought this name indicates that the testplan is for the specific image.

skuusela:
This can actually be merged as it only creates a new manifest file and shouldn't break anything.

skuusela:
test this please

okartau:
test this please

PR #12, opened 2017-02-10
CI Jenkinsfile: update GIT settings for CI user

okartau:
Publishing build results over rsync does not gain anything from being
run inside docker, but data path gets more complex and more risky
to introduce overhead and problems, so letäs reduce that complexity.


okartau:
test this please

okartau:
test this please

PR #15, opened 2017-02-10
openssh_%.bbappend: fix incorrect iptables dependency.

klihub:
The openssh base package is empty. Having the iptables runtime
dependency there is useless, as the base package never gets
installed. Move the dependency to the proper -sshd subpackage.


klihub:
That test failure is a weird CI/QA error, unrelated to this change...

PR #18, opened 2017-02-13
Add librealsense to computer vision profile.

ipuustin:
Add `meta-intel-realsense` layer. Fix librealsense recipe so that it compiles without X11 support. Patch librealsense itself so that we can drop bash runtime dependency and have non-graphical examples for testing and trying out librealsense. The local patches have been submitted to upstream, so in the future we might drop the three patches that change librealsense behavior.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,3 +25,6 @@
 [submodule "meta-security-isafw"]
 	path = meta-security-isafw
 	url = https://github.com/01org/meta-security-isafw.git
+[submodule "meta-intel-realsense"]
+	path = meta-intel-realsense
+	url = https://github.com/IntelRealSense/meta-intel-realsense.git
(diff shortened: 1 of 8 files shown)


meta-refkit/recipes-computervision/librealsense/librealsense_1.12.0.bbappend
```diff
@@ -7,3 +9,8 @@ EXTRA_OECMAKE = " \
  
 PACKAGES = "${PN} ${PN}-dbg ${PN}-dev ${@bb.utils.contains('DISTRO_FEATURES', 'x11 opengl', '${PN}-examples ${PN}-graphical-examples', '', d)}"
 
+SRC_URI_append = " \
+    file://0001-scripts-removed-bashisms.patch \
```
mythi: Is it expected these patches are fro 1.12.0 only? When IntelRealSense/meta-intel-realsense/pull/8 gets merged and updated here the patches are no longer appended.

ipuustin: You're right... This should work even with a wildcard version match after the version update. Changing the patch set.

ipuustin:
Updated patches and rebased.

PR #19, opened 2017-02-14
[TEST BUILD] Installer image

pohly:
Includes also support for whole-disk encryption and dm-verity (currently limited to the installer image).

Whole-disk encryption depends on TPM 1.2 and recipes from meta-security with patches that are currently pending in meta-security master-next (for example, "trousers: tcsd.conf must be owned tss:tss"). The PR can be merged, but the feature can't be enabled.

There are some TODOs that depend on extending the product vs. development image concept. There was a discussion around that also in OE-core (triggered by the proposal to set a root password by default); we should reach some conclusion for OE-core and used that for our refkit-production.inc and 
refkit-development.inc files and also these TODO items. For now I propose to merge as-is.

Documentation only exists in the form of source code comments.


pohly:
Question: should refkit-installer-image be created by default?

pohly:
I suppose we want the installer image, so I've enabled it also by default.

pohly:
Building multipath-tools fails - a known issue, the patch fixing it just isn't in refkit yet.

pohly:
test this please

I've added the updated meta-openembedded with the multipath-tools fix.

meta-refkit/conf/distro/refkit.conf
```diff
@@ -307,3 +307,7 @@ INHERIT += "uninative"
 # stage due to lack of RTC backup battery. This causes unnecessary fixes being
 # made due to filesystem metadata time stamps being in future.
 APPEND_append = " fsck.mode=skip"
+
+KERNEL_FEATURES_append = "${@ bb.utils.contains('DISTRO_FEATURES', 'dm-verity', ' dm-verity.scc', '', d) }"
+SRC_URI_append_pn-linux-yocto = " file://dm-verity.scc"
+FILESEXTRAPATHS_prepend_pn-linux-yocto = "${META_REFKIT_BASE}/files/linux-yocto:"
```
mythi: @pohly linux-intel is what we get from meta-intel nowadays. Have you submitted these in yocto-kernel-cache already? We've targeted to not have BSP bbappends here but all comes from meta-intel.

pohly: Yes, I have submitted it, but it's not yet available. The entire commit can be reverted once it is.

pohly:
Please do not merge quite yet. While re-testing the dm-verity change without the local.conf changes that I was using earlier I am running into some problem with the verity.scc that I need to investigate first.

meta-refkit/conf/distro/refkit.conf
pohly: Similar modifications for the new linux-intel need to be added.

mythi: It uses the same kernel cache. Once the scc is merged there let's submit a meta-intel patch ourselves to speed up things.

meta-refkit/classes/refkit-image.bbclass
```diff
@@ -274,7 +274,11 @@ IMAGE_FSTYPES_remove = "live"
 # Activate "dsk" image type.
 IMAGE_CLASSES += "${@ 'image-dsk' if ${REFKIT_USE_DSK_IMAGES} else ''}"
 
+# By default, the full image is going to use roughly 4GB, independent
+# of the actual roofs size.
 WKS_FILE = "refkit-directdisk.wks.in"
+REFKIT_VFAT_MB ??= "30"
```
jlaako: I recommend using 32

meta-refkit/classes/refkit-image.bbclass
```diff
@@ -274,7 +274,11 @@ IMAGE_FSTYPES_remove = "live"
...
+REFKIT_IMAGE_SIZE ??= "--fixed-size 3700M"
```
jlaako: And 3968 here

PR #20, opened 2017-02-14
meta-iotqa: Enable basic Bluetooth test and add masking tests

skuusela:


skuusela:
The bluetooth tests didn't seem to execute on the Joule. The reason seems to be that on my test setup I have newer BIOS version with Joule and `cat /sys/devices/virtual/dmi/id/product_name` returns "570x ..." and on CI "Broxton M".  So this way of identifying the hardware and if the bluetooth test should be run is a bit fragile. @mythi Thoughts? Maybe add both "Broxton" and "570x" to the list of devices with bluetooth so the test will execute now and also later if we update the BIOS.

mythi:
@skuusela I feel the logic to decide which cases are not compatible with certain boards should be done outside the test cases. Could it be some <device>.mask file with a list of regexps about incompatible tests. that mask can be applied in tester-exec.sh before calling DAFT.

skuusela:
@mythi Better? Not sure if this was the way you were thinking.

mythi:
Yes, this is what I was thinking.  looks good to me

mythi:
test this please

mythi:
Well, one concern though. This way we loose statistics of how many tests are skipped. Should be doable to update the statistics based on masking results though.

I'd still prefer the masking like this because it allows to keep the test cases hardware agnostic.

skuusela:
@mythi I'm not sure if it's that useful to know if we skip tests. For example it's not surprising to not run Bluetooth tests on Minnowboard as it doesn't have it. One possibility could also be when making the image-testplan.manifest to make one for the Joule and one for the Minnowboard eg. 570x-image-testplan.manifest and minnowboardturbot-image-testplan.manifest. I could then configure DAFT to use the corresponding test manifest depending on the DUT. Maybe that would be the cleanest way?

mythi:
@skuusela masking some tests is skipping IMO. I think it *is* useful to know which tests from an image testplan are skipped for certain HW. We may some day buy a Minnowmax Lure and connect an mPCI card with wifi/BT to minnowboard and get the tests running -> easy to unmask those tests.

skuusela:
test this please

skuusela:
@mythi Now the masked tests will be counted as skipped tests in the final report. Aren' the mask files enough to keep the info about tests that are skipped for certain hardware?

docker/tester-exec.sh
```diff
@@ -93,6 +93,14 @@ testimg() {
     RECORD_ARG=""
   fi
 
+  # Remove incompatible tests for the DUT from image-testplan.manifest
+  MASKFILE=./iottest/testplan/${DEVICE}.mask
+  MANIFEST=./iottest/testplan/image-testplan.manifest
+  set +e # 'grep -cFwf' exit code isn't 0 if no tests get masked so ignore it
+  num_masked=$(grep -cFwf ${MASKFILE} ${MANIFEST})
```
mythi: Looks good to me. @okartau +1?

PR #21, opened 2017-02-15
Pulseaudio: enable the use of pulseaudio packages

juimonen:
Add pulseaudio to distro features and supported
recipes. Create also audio group and pulse user.
This commit doesn't include any pulseaudio packages
to the images.

### DIFF
--- a/meta-refkit/conf/distro/include/refkit-supported-recipes.txt
+++ b/meta-refkit/conf/distro/include/refkit-supported-recipes.txt
@@ -31,6 +31,7 @@
 
 acl@core
 alsa-lib@core
+alsa-plugins@core
 alsa-state@core
 alsa-utils@core
 appfw-test-app@iotqa
@@ -221,6 +222,7 @@ ppp@core
 procps@core
 protobuf@openembedded-layer
 ptest-runner@core
+pulseaudio@core
 python-dbus@core
 python-nose@core
 python-numpy@core
@@ -239,6 +241,7 @@ rmc-db@intel
 rmc@intel
 rsync@core
 run-postinsts@core
+sbc@core
 sed@core
 shadow-securetty@core
 shadow@core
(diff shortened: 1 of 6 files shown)


mythi:
@juimonen avahi is one of the pulseaudio ```PACKAGECONFIG```. We'd either need to drop that or add avahi in supported recipes and add the users it needs. I'd suggest we start without avahi in the ```PACKAGECONFIG```

juimonen:
ok, i forgot add the avahi removal for this pr. I can remove the speex at the same time

meta-refkit/recipes-multimedia/alsa/alsa-plugins_%.bbappend
```diff
@@ -0,0 +1 @@
+PACKAGECONFIG_remove = "speexdsp"
```
mythi: move to refkit.conf (see, e.g., gstreamer-plugins-base).

meta-refkit/recipes-multimedia/pulseaudio/pulseaudio_%.bbappend
```diff
@@ -0,0 +1,4 @@
+DEPENDS_remove = "speexdsp"
+DEPENDS += "libsamplerate0"
+PACKAGECONFIG_remove = "avahi"
```
mythi: ditto here. ```--disable-avahi``` is redundant after the packageconfig change

meta-refkit/conf/distro/include/refkit-supported-recipes.txt
```diff
@@ -84,6 +85,7 @@ freetype@core
 gawk@core
 gcc-runtime@core
 gcc@core
+gconf@core
```
jlaako: Does pulseaudio itself really need gconf?

juimonen: it is required by some module that is included in the oe-recipe. I think it can be disabled from configure. Can try that tomorrow...

juimonen:
removed gconf

mythi:
test this please

PR #22, opened 2017-02-15
meta-iotqa: Ignore Minnowboard boot error

skuusela:
Add '\*ERROR\* dp aux hw did not signal timeout' to list of ignored
errors. This error happens randomly under 5% of the time and doesn't
seem to cause any additional errors in other tests. Tried to reproduce
the error with an image that it happened with but couldn't get it with 25
flash/test cycles. This error also doesn't seem to be easily fixed. So ignore this
error as it will cause tests to fail randomly.

### DIFF
--- a/meta-iotqa/lib/oeqa/runtime/sanity/baseos.py
+++ b/meta-iotqa/lib/oeqa/runtime/sanity/baseos.py
@@ -113,7 +113,8 @@ class BaseOsTest(oeRuntimeTest):
             "Direct firmware load for iwlwifi-8000C",
             "ACPI Error: Could not enable RealTimeClock event",
             "hci_intel: probe of INT33E1:00 failed with error -2",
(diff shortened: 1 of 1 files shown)


okartau:
We should not mask something without study, explanation or documenting it, just to get tests through.
There is an open Freedesktop issue documenting a similar symptom:
https://bugs.freedesktop.org/show_bug.cgi?id=96428

skuusela:
@okartau Yes would be nice if these errors we ignore could be fixed or documented somewhere. I also found lots of posts about this error and it seems to occur with lots of different hardware and distros and also didn't really find anything useful that could fix this for sure. This error is still quite rare, doesn't seem to affect anything and at the moment we are just retriggering the testing again if a PR fails. Usually the PR succeeds the next time. So I would just ignore this error.

PR #29, opened 2017-02-16
Add bluetooth audio packages to gateway profile

juimonen:
Add pulseaudio and its bluetooth related modules
to gateway profile. Includes also a simple test to
check the existence of the daemon.

### DIFF
--- a/meta-iotqa/lib/oeqa/runtime/sanity/pulseaudio.py
+++ b/meta-iotqa/lib/oeqa/runtime/sanity/pulseaudio.py
@@ -1,0 +1,11 @@
+from oeqa.oetest import oeRuntimeTest
+from oeqa.utils.decorators import tag
+
+class PulseaudioTest(oeRuntimeTest):
+
+    '''Check pulseaudio existence'''
+    def test_pulseaudio_exists(self):
+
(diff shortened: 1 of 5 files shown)


meta-iotqa/lib/oeqa/runtime/sanity/pulseaudio.py
```diff
@@ -0,0 +1,12 @@
+from oeqa.oetest import oeRuntimeTest
+from oeqa.utils.decorators import tag
+
+class PulseaudioDaemonTest(oeRuntimeTest):
+
+    '''Pulseaudio daemon check'''
+    def test_pulseaudio_daemoncheck(self):
+        '''check pulseaudio daemon'''
```
skuusela: Maybe rename the the class as 'PulseaudioTest' and the test function as 'test_pulseaudio_exists' and also the comments to reflect those.

juimonen: ok done.

mythi:
the oe-core update is not needed. Also, for better granularity this needs three commits: add the feature, add the tests, enable the feature in gateway

juimonen:
ok. tried to do the commit split.

PR #31, opened 2017-02-17
meta-iotqa: Add REST API server checking and stability tests

qiuzhong:
* Add 12 test cases for checking if REST API server returns duplicate entries of same
OCF devices/platforms/resources in the discovered list
* Add 2 stability test cases for the REST API server.

Tested on Joule with IoT RefKit image#107
New:14 Pass:14 Failed:0

Signed-off-by: Zhong Qiu <zhongx.qiu@intel.com>

### DIFF
--- a/meta-iotqa/conf/test/restapichecklocal.manifest
+++ b/meta-iotqa/conf/test/restapichecklocal.manifest
@@ -1,0 +1,1 @@
+oeqa.runtime.nodejs.rest_check_local
--- a/meta-iotqa/conf/test/restapicheckremote.manifest
+++ b/meta-iotqa/conf/test/restapicheckremote.manifest
@@ -1,0 +1,2 @@
+oeqa.runtime.nodejs.rest_one_ocf_server_remote
+oeqa.runtime.nodejs.rest_multi_ocf_servers_remote
--- a/meta-iotqa/conf/test/stability.manifest
+++ b/meta-iotqa/conf/test/stability.manifest
@@ -1,0 +1,1 @@
+oeqa.runtime.nodejs.rest_stability
--- a/meta-iotqa/lib/oeqa/runtime/nodejs/copy_necessary_files.py
+++ b/meta-iotqa/lib/oeqa/runtime/nodejs/copy_necessary_files.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+
+import os
+
+def copy_smarthome_demo_ocf_server(ip, to_dir = '/home/root'):
+	'''
+	Copy SmartHome-Demo/ocf-servers directory to target IP.
+	'''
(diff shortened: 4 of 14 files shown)


mythi:
A couple of comments and questions about the test cases:

- do they depend on the smarthome  demo peripherals (LED/gas) connected to the target device?
- the long stability test cases are not suitable for a CI environment we'd expect these cases to run

wanghongjuan:
Hi, @mythi Thanks for your review and comment.
- For the first question: these test cases are not depend on the smarthome demo, we just need to use codes of the LED/gas to create resources, and we will adjust the code to avoid doubt like you mentioned.
- Second, I agree your point,but I think iot-rest-api-server need to use these test cases to check if it can work well. Is there any solution can fix this problem? Can we make a config list to specify manifest files which are suitable for CI environment?

wanghongjuan:
We updated the code. @mythi @nagineni Could you please take a look?

skuusela:
If these tests are added, why not put them on some profiles manifest so they are actually used and tested that they work on our setup? So maybe put these to refkit-image-common.manifest, refkit-image-gateway.manifest or refkit-image-computervision.manifest.

wanghongjuan:
Hi, @skuusela Thanks for your suggestion. I noticed that the CI environment seems like only run tests which are setting in the three profiles manifest that you mentioned, most of them are sanity tests for feature verification, how about extension tests for feature verification, like [wifi](https://github.com/intel/intel-iot-refkit/tree/master/meta-iotqa/lib/oeqa/runtime/wifi)? There is not manifest file for them. Our tests are similar to wifi. I think these tests are no necessary add to CI Environment, especially stability test, they are not suitable CI, we only need to run it when iot-rest-api-server has new release version.

mythi:
On Wed, Feb 22, 2017 at 8:54 AM, wanghongjuan wrote:

> for them. Our tests are similar to wifi. I think these tests are no
> necessary add to CI Environment, especially stability test, they are not
> suitable CI, we only need to run it when iot-rest-api-server has new
> release version.
>

I'd suggest the following: maintain these test cases in meta-iot-web. It'd
still be possible for you
to use them in release testing with a combination of intel-iot-refkit +
meta-iot-web (with updates).

Additional review comments:
- can you avoid the ```copy_necessary_files.py``` by making the assumption
that the necessary smarthome-* recipes are installed in the image? Also,
I'm thinking meta-iot-web could add a recipe (similar to johnny-five) to
install mocha.
- do you need to run as root or can you use ```restful``` that is created
for iot-rest-api-server?

If you agree meta-iot-web is an OK place for these tests, I'd close this PR.

wanghongjuan:
@mythi I agree with you. For you review comments, we will adjust our code, we are run these test as root, and close this PR.

PR #34, opened 2017-02-20
Remove explicit EFI subfolder copy when creating a FAT partition

jlaako:
In order to properly copy rmc.db and other EFI partition contents,
remove explicit EFI subfolder copy and instead copy all boot/ contents
to the EFI partition. Also make the EFI partitions larger to properly
fit all the contents and the total image size 64 MB under 4 GB.

### DIFF
--- a/meta-refkit/lib/image-dsk.py
+++ b/meta-refkit/lib/image-dsk.py
@@ -44,7 +44,9 @@ def populate_rawcopy(src, dst):
 def populate_vfat(src, dst):
     """Create and populate a FAT partition, out of a root directory <src>."""
     check_call(['mkdosfs', dst])
-    check_call(['mcopy', '-i', dst, '-s', src + '/EFI', '::/'])
+    """this requires shell expansion for wildcard
+    check_call(['mcopy', '-i', dst, '-s', src + '/*', '::/'])"""
+    os.system('mcopy -i ' + dst + ' -s ' + src + '/* ::/')
 
 
 def populate_ext4(src, dst):


mythi:
@jlaako we dont't build the dsk images anymore so the rmc db copying problem was not seen.  Is the wic image not suitable for you?

jlaako:
@mythi I am building dsk images because I don't know why I would want wic images and how to use those. But anyway your patch for rmc was changing this dsk image recipe so I thought it was supposed to work. And IMO as long as there is possibility to build dsk/tar/whatever type of images, the correct things should be copied to correct places. If CI doesn't cover all the image types, then it's a CI deficit.

mythi:
@jlaako The wic images are used just like dsk. It's recommended to move to the wic images as dsk is going away soon once I find the final home for ```do_uefiapp```.

This fix looks OK until then. However, I wonder how it used to work with, e.g, emmc_partitions_data...

About the partition sizes. Can you split that into another commit and align the settings with refkit-directdisk.wks.in too

jlaako:
@mythi I'll split the partition size part out and sync.

jlaako:
This should be preferably merged together with https://github.com/intel/intel-iot-refkit/pull/36 otherwise dsk image creation will fail due to insufficient space.

pohly:
On Tue, 2017-02-21 at 02:27 -0800, Jussi Laako wrote:
> @mythi I'll split the partition size part out and sync.

I ran into the same issue as part of the image installer work.
See
https://github.com/intel/intel-iot-refkit/pull/19/commits/ecae57186602f75b72e11de08c1335ae94970d12

Can we perhaps merge that instead? It is a bit more flexible.

Shall I submit a PR with it or can Jussi update his PR with that commit?

pohly:
On Tue, 2017-02-21 at 01:22 -0800, Mikko Ylinen wrote:
> This fix looks OK until then. However, I wonder how it used to work
> with, e.g, emmc_partitions_data...

The installed VFAT partition was smaller than the one used in the
images. I noticed when copying bootx64.efi during installation failed
once it became bigger.

mythi:
@pohly ecae571 looks OK but let's change the sizes based on @jlaako's feedback. Will you submit a PR for it?

PR #39, opened 2017-02-22
Enable and clean up wifi connect test

skuusela:
Test connecting to a wifi access point. Also add 'common-test' features to all development profile images. This way all images will have connman-client and wifi can be tested with all images.


skuusela:
Wifi test seem to run fine but the duration of 170s to just connect to a wifi AP is a bit long. Will investigate why it takes so long.

skuusela:
Now it takes 20s. I removed one 30s sleep after enabling wifi. The comment said it's used because when enabling wifi it automatically connects to last AP. I tested this manually and it actually didn't seem to work that way and the tests worked fine whitout the timeout multiple times.

skuusela:
test this please

skuusela:
test this please

PR #42, opened 2017-02-23
meta-iotqa: Ignore one alsa boot error

skuusela:
The error: "/usr/sbin/alsactl: state_lock:125: file
/var/lib/alsa/asound.state lock error: File exists" happens randomly under 2% of
the time so it's really hard to replicate and test if it affects
anything. Also there seems to be no fix for it so ignore the error.

### DIFF
--- a/meta-iotqa/lib/oeqa/runtime/sanity/baseos.py
+++ b/meta-iotqa/lib/oeqa/runtime/sanity/baseos.py
@@ -106,7 +106,9 @@ class BaseOsTest(oeRuntimeTest):
             "ACPI Error: Could not enable RealTimeClock event",
             "hci_intel: probe of INT33E1:00 failed with error -2",
             "Error changing net interface name 'usb0' to",
-            "*ERROR* dp aux hw did not signal timeout"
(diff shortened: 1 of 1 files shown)


okartau:
test this please

mythi:
I had not seen this. It smells like a race which *are* hard to replicate. Is this with both 570x and minnowboard?

skuusela:
@mythi I can only see this error one time in recent history with Joule but I think it happens with both.

ipuustin:
I think it would be better to try to find the bug. Let's wait for @juimonen to comment.

skuusela:
@ipuustin I actually asked @juimonen week or two ago about this and he couln't tell anything more about it  neither.

mythi:
I'd propose from now on (we have bugzilla) we don't ignore errors until a bug is filed. With that the problems don't get forgotten + we get to triage the priority of fixing the problem.

The bug ID could be added in the log

skuusela:
@mythi Submitted a bug to Refkit Bugzilla and added it's id as comment to the ignored error.

PR #44, opened 2017-02-23
image: adjust partition sizes

pohly:
They were inconsistent and too small for future use cases.

Replaces PR #36 and needed for PR #19, which already had commit
13829d8.

### DIFF
--- a/meta-refkit/classes/image-dsk.bbclass
+++ b/meta-refkit/classes/image-dsk.bbclass
@@ -75,7 +75,7 @@ DSK_IMAGE_LAYOUT ??= ' \
     "partition_01_primary_uefi_boot": { \
         "name": "primary_uefi", \
         "uuid": 0, \
-        "size_mb": 15, \
+        "size_mb": ${REFKIT_VFAT_MB}, \
         "source": "${IMAGE_ROOTFS}/boot/", \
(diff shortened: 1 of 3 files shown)


mythi:
@pohly are the submodule updates necessary?

pohly:
On Thu, 2017-02-23 at 07:28 -0800, Mikko Ylinen wrote:
> @pohly are the submodule updates necessary?

No, that crept in unintentionally (juggling between my image-installer
branch and this PR). I'll amend it.

pohly:
Force-pushed updated commit.

mythi:
test this please

ipuustin:
@pohly I also keep committing the submodules unintentionally (`git commit .` is the main cause). I was thinking of a commit hook to warn about it, but I don't know if that would interfere too much with the legitimate submodule updates.

pohly:
@ipuustin if the change for the submodule was a bit more obvious in "git diff" output perhaps it would be easier to spot the mistake - at least I tend to check my change that way before committing. But I've not found a way to get more than the two-line change with old and new hash, which is just too easy to miss.

mythi:
+1

@jlaako Are you OK?

jlaako:
@mythi yes, +1

PR #45, opened 2017-02-24
Add and enable mraa GPIO tests

skuusela:
These tests will set a GPIO pin on and off and check the output with BeagleBone that it works.

### DIFF
--- a/meta-iotqa/lib/oeqa/runtime/sanity/mraa_gpio.py
+++ b/meta-iotqa/lib/oeqa/runtime/sanity/mraa_gpio.py
@@ -1,0 +1,44 @@
+from oeqa.oetest import oeRuntimeTest
+import unittest
+import subprocess
+from time import sleep
+
+class MraaGpioTest(oeRuntimeTest):
+    '''
+    These tests require to use BeagleBone as testing host
+    '''
+    pin = ""
+    def setUp(self):
+        (status, output)= self.target.run("mraa-gpio version")
+        output = output.lower()
+        if any(x in output for x in ("broxton", "tuchuck", "joule")):
+            self.pin = "51"
+        elif "minnowboard" in output:
+            self.pin = "25"
+        else:
+            raise unittest.SkipTest(output)
+
+    def test_gpio(self):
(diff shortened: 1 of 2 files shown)


skuusela:
test this please

skuusela:
@mythi This is kinda the same problem as with adding Bluetooth test but this test has to be hardware specific. I also made alternative way of doing this without using the mask files: https://github.com/skuusela/intel-iot-refkit/commit/59f80e77aafc491ead495de7f1dbacb9e7b07211.

mythi:
How reliable is mraa's capability to detect the board? See ```mraa-gpio version```

skuusela:
@mythi On Joule it gives: "Version v1.5.1 on Intel GT Tuchuck" and on Minnowboard Turbot: "Version v1.5.1 on MinnowBoard MAX"

mythi:
Oh, I did not actually check https://github.com/skuusela/intel-iot-refkit/commit/59f80e77aafc491ead495de7f1dbacb9e7b07211 before my guestion. I'd prefer that option because it also tests mraa's capability to detect the board at the same time.

mythi:
test this please

meta-iotqa/lib/oeqa/runtime/sanity/mraa_gpio.py
```diff
@@ -0,0 +1,44 @@
...
+        else:
+            raise unittest.SkipTest(output)
+
+    def test_gpio(self):
+        '''
+        Test a GPIO pin on and off and check the pin output with
+        BeagleBone
+        '''
+        def check_gpio_output():
+            cmd = "cat /sys/class/gpio/gpio20/value".split()
```
mythi: can you do this directly without exporting the GPIO first?

skuusela: No the GPIO pin has to be exported, but it's already exported on the Beaglebones. I added few lines to a boot script on the Beaglebone so it's always exported and ready after booting.

PR #50, opened 2017-02-27
RMC hardening + NUC boot fixes

mythi:


mythi:
@pohly FYI

pohly:
Where is the code which used to read the RMC DB from the VFAT system partition? Is that now gone for sure?

mythi:
@pohly That's in ```SRC_URI_remove_intel-x86-common = "file://0004-sd-boot-Support-global-kernel-command-line-fragment-in-EFI-stub.patch"```

pohly:
On Tue, 2017-02-28 at 00:03 -0800, Mikko Ylinen wrote:
> @pohly That's in SRC_URI_remove_intel-x86-common =
> "file://0004-sd-boot-Support-global-kernel-command-line-fragment-in-EFI-stub.patch"

I hadn't considered that meta-intel carries that patch. This
SRC_URI_remove is a bit fragile - consider what happens when the patch
gets renamed (for whatever reason) in meta-intel...

I'm also a bit concerned that we now have systemd-boot patches in both
meta-intel and meta-refkit. I understand that this is necessary because
the UEFI app currently is a refkit-only thing. Perhaps now would be a
good opportunity to move it to meta-intel, together with these
enhancements?

mythi:
@pohly this is just a test-build. I'm going to drop fc6803a once meta-intel has those patches. They are already submitted there.

pohly:
On Tue, 2017-02-28 at 02:36 -0800, Mikko Ylinen wrote:
> @pohly this is just a test-build. I'm going to drop fc6803a once
> meta-intel has those patches. They are already submitted there.

Makes sense. That it's just a test build wasn't clear.

mythi:
not serving a purpose now

PR #53, opened 2017-02-28
CI: run oe-selftests as pre-build, post-build steps

okartau:
First 5 commits are re-factoring of Jenkinsfile, as I prepared to introduce separate stage where prebuild
would run. 
After that done, I was advised to insert prebuild test in build script. 
That was done in 6th commit.
Commit 7 drops use of ssh-agent on docker session as we seem to get repos accessed without it.
Then the plan changed again and pre- and -post steps were added as separate stages and scripts
in commits 8,9 with build step un-parallelised (one MACHINE).
This restored some stages taken away in refactoring.
Commit 11 clarifies build script parts by moving some more sub-steps to CI-only sections, freeing
local build from CI and tester related steps.
last commit 12 moves enablement of BUILDHISTORY from refkit-ci.inc into build-project.sh, thus freeing
refkit-ci.inc to be used in oe-selftest runs, taking away need to overwrite auto.conf, and making pre-build and post-build scripts simpler. Note that post-build script has to reset auto.conf anyway because build phase inserted BUILDHISTORY setting there. But now the auto.conf can be rewritten using refkit-ci.inc

### DIFF
--- a/docker/build-project.sh
+++ b/docker/build-project.sh
@@ -80,6 +80,9 @@ EOF
   fi
   # Buildhistory mode set always in CI run
   cat >> conf/auto.conf << EOF
+INHERIT += "buildhistory"
+BUILDHISTORY_COMMIT = "1"
+INHERIT += "buildhistory-extra"
 BUILDHISTORY_DIR ?= "${BUILDHISTORY_TMP}"
 EOF
   if [ ! -z ${COORD_BASE_URL+x} ]; then
--- a/docker/build-project.sh
+++ b/docker/build-project.sh
@@ -25,9 +25,6 @@ cd $WORKSPACE
 CI_GIT_COMMIT=$(git rev-parse HEAD)
 echo ${CI_GIT_COMMIT} > ci_git_commit
 
-# document env.vars in build log
-env |sort
-
 # use +u to avoid exit caused by unbound variables use in init scripts
 set +u
 # note, BUILD_DIR is also undef in CI case, but is set in local-build case.
@@ -55,7 +52,7 @@ if [ ! -z ${JOB_NAME+x} ]; then
   popd
 fi
 
-# Take local CI preferences if present
+# Initialize auto.conf from local CI preferences if present
 if [ -f $WORKSPACE/meta-*/conf/distro/include/refkit-ci.inc ]; then
   cat $WORKSPACE/meta-*/conf/distro/include/refkit-ci.inc > conf/auto.conf
 fi
@@ -96,9 +93,14 @@ export BUILD_ID=${CI_BUILD_ID}
 export BB_ENV_EXTRAWHITE="$BB_ENV_EXTRAWHITE BUILD_ID"
 
 if [ -z "$BUILD_TARGET" ]; then
-  # Let's try to fetch build targets from configuration files
-  bitbake -e >bb_e_out 2>bb_e_err || (cat bb_e_err && false)
-  grep -E "^REFKIT_CI" bb_e_out > ${WORKSPACE}/refkit_ci_vars || true
+  # Let's try to fetch build targets from configured variables.
+  # refkit_ci_vars is made in pre-build script in CI run, but
+  # may need to be created in local-build run.
+  if [ ! -f ${WORKSPACE}/refkit_ci_vars ]; then
+    # use bitbake -e for variables parsing, then pick REFKIT_CI part
+    bitbake -e >bb_e_out 2>bb_e_err || (cat bb_e_err && false)
+    grep -E "^REFKIT_CI" bb_e_out > ${WORKSPACE}/refkit_ci_vars || true
+  fi
   _bitbake_targets=""
   for ci_var in `perl -pe "s/^([A-Z_]+)=.+/\1/g" ${WORKSPACE}/refkit_ci_vars`; do
     case "$ci_var" in
@@ -133,7 +135,7 @@ fi
 # Push buildhistory into machine-specific branch in the master buildhistory
 #
 if [ ! -z ${JOB_NAME+x} ]; then
-  cd ${BUILDHISTORY_TMP}
+  pushd ${BUILDHISTORY_TMP}
   BUILDHISTORY_TAG="${JOB_NAME}/${CI_BUILD_ID}/${CI_GIT_COMMIT}/${TARGET_MACHINE}"
   git tag -a -m "Build #${BUILD_NUMBER} (${BUILD_TIMESTAMP}) of ${JOB_NAME} for ${TARGET_MACHINE}" -m "Built from Git revision ${CI_GIT_COMMIT}" ${BUILDHISTORY_TAG} refs/heads/${BUILDHISTORY_BRANCH}
 
@@ -141,6 +143,7 @@ if [ ! -z ${JOB_NAME+x} ]; then
   # push branch might fail if multiple concurent jobs running for this branch.
   # That's ok, as most important part is stored under tag.
   git push origin refs/heads/${BUILDHISTORY_BRANCH}:refs/heads/${BUILDHISTORY_BRANCH} || true
+  popd
 fi
 
 # #############
--- a/docker/build-project.sh
+++ b/docker/build-project.sh
@@ -21,25 +21,26 @@ BUILD_TARGET="$@"
 export LC_ALL=en_US.UTF-8
 
 cd $WORKSPACE
-# get git commit ID of the project for using in buildhistory tagging
-CI_GIT_COMMIT=$(git rev-parse HEAD)
-echo ${CI_GIT_COMMIT} > ci_git_commit
 
 # use +u to avoid exit caused by unbound variables use in init scripts
 set +u
-# note, BUILD_DIR is also undef in CI case, but is set in local-build case.
+# BUILD_DIR is undef in CI case, but is set in local-build case.
 source refkit-init-build-env ${BUILD_DIR}
 set -u
 
 if [ ! -z ${JOB_NAME+x} ]; then
-  # Prepare for buildhistory generation
-  # If JOB_NAME is defined, we commit buildhistory in
-  # BH branch, name of which is composed from JOB_NAME and MACHINE.
+  # in CI run only:
+  # get commit ID for using in buildhistory tagging, save for later use
+  CI_GIT_COMMIT=$(git rev-parse HEAD)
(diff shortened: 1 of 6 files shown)


okartau:
test this please

okartau:
test this please

okartau:
test this please

okartau:
test this please

okartau:
test this please

okartau:
test this please

okartau:
test this please

okartau:
test this please

okartau:
test this please

okartau:
test this please

okartau:
test this please

okartau:
test this please

okartau:
test this please

docker/build-project.sh
```diff
@@ -34,7 +34,23 @@ set +u
...
+
+# use bitbake -e for variables parsing, then pick REFKIT_CI part
+bitbake -e >bb_e_out 2>bb_e_err || (cat bb_e_err && false)
+grep -E "^REFKIT_CI" bb_e_out > ${WORKSPACE}/refkit_ci_vars || true
+
 if [ ! -z ${JOB_NAME+x} ]; then
+  # in CI: run pre-build oe-selftests using development-mode settings in auto.conf
+  echo "include conf/distro/include/refkit-development.inc" > ${WORKSPACE}/build/conf/auto.conf
+  for _test in `grep REFKIT_CI_PREBUILD_SELFTESTS ${WORKSPACE}/refkit_ci_vars | perl -pe 's/.+="(.*)"/\1/g; s/[^ .a-zA-Z0-9_-]//g'`; do
+      oe-selftest --run-tests ${_test}
```
pohly: I find it nicer to invoke oe-selftest only once, with the entire set of tests as parameter.

docker/build-project.sh
```diff
@@ -34,7 +34,23 @@ set +u
...
+  cat $WORKSPACE/meta-*/conf/distro/include/refkit-ci.inc > conf/auto.conf
+fi
+
+# use bitbake -e for variables parsing, then pick REFKIT_CI part
+bitbake -e >bb_e_out 2>bb_e_err || (cat bb_e_err && false)
+grep -E "^REFKIT_CI" bb_e_out > ${WORKSPACE}/refkit_ci_vars || true
+
 if [ ! -z ${JOB_NAME+x} ]; then
+  # in CI: run pre-build oe-selftests using development-mode settings in auto.conf
+  echo "include conf/distro/include/refkit-development.inc" > ${WORKSPACE}/build/conf/auto.conf
```
pohly: An explanation that this can't use the full refkit-ci.inc because of the buildhistory issue would be useful, as it explains why the script goes to all the trouble of rewriting auto.conf multiple times.

docker/build-project.sh
```diff
@@ -133,14 +147,21 @@ fi
...
   git push origin refs/tags/${BUILDHISTORY_TAG}:refs/tags/${BUILDHISTORY_TAG}
   # push branch might fail if multiple concurent jobs running for this branch.
   # That's ok, as most important part is stored under tag.
   git push origin refs/heads/${BUILDHISTORY_BRANCH}:refs/heads/${BUILDHISTORY_BRANCH} || true
+  popd
+
+  # in CI: run post-build oe-selftests using development-mode settings in auto.conf
+  echo "include conf/distro/include/refkit-development.inc" > ${WORKSPACE}/build/conf/auto.conf
+  for _test in `grep REFKIT_CI_POSTBUILD_SELFTESTS ${WORKSPACE}/refkit_ci_vars | perl -pe 's/.+="(.*)"/\1/g; s/[^ .a-zA-Z0-9_-]//g'`; do
+      oe-selftest --run-tests ${_test}
```
pohly: Same here - better only run oe-selftest once.

meta-refkit/conf/distro/include/refkit-ci.inc
```diff
@@ -49,6 +49,11 @@ require conf/distro/include/refkit-development.inc
 REFKIT_VM_IMAGE_TYPES = "wic.xz wic.zip wic.bmap wic.xz.sha256sum"
 
 #
+# pre/post-build oe-selftests started by CI builder, whitespace-separated.
```
pohly: Can you add a note here explaining that refkit-ci.inc itself is not active while running the selftests and why?

meta-refkit/lib/oeqa/selftest/iotsstatetests.py
```diff
@@ -33,7 +33,8 @@ def test_sstate_samesigs(self):
         libcappend = get_bb_var('TCLIBCAPPEND')
         # Select subset of the machines to speed up testing.
         # Edison/intel-core2-32 are particularly sensitive.
-        machines = "edison intel-quark intel-core2-32 intel-corei7-64 beaglebone".split()
+        machines = "intel-corei7-64".split()
```
okartau: I asked in bugz. ticket does it make sense to enable this test in refkit, and you replied "in principle it might still be useful", so I guessed there may be something usable even with one machine. But other than that, yes, it's used as activation trial now, plus it will be ready to strike after someone adds more MACHINE targets.

meta-refkit/conf/distro/include/refkit-ci.inc
okartau: I added one more comment line there.
Does the current version address all change requests for now?

pohly: Yes, I think it is good enough for now. Long-term I suspect that we might have a need to configure additional settings for oe-selftest, for example to make it more similar to the actual build. But that can wait.

If that becomes necessary, we could have a refkit-ci.inc and a refkit-selftest.inc where refkit-ci.inc includes refkit-selftest.inc and adds buildhistory. That's also a bit cleaner than the current "include refkit-development.inc" (which so far was an implementation detail of the distro and invisible to the build system).

mythi:
c5f3ee7 commit message looks misleading as it does not touch Jenkinsfile. You could also add the bug number in the message.

I was not able to find the arguments for doing pre-/post- testing in the build-project.sh script instead of their own jenkins steps. With this approach we get all selftest failures as "build failures". 

btw, what is the problem with buildhistory + oe-selftests?

okartau:
the arguments for doing pre-/post- testing in the build-project.sh script instead of their own jenkins steps: I got guidance from bugz.ticket 11111 comments where pohly expressed his vision that tests should be run in same context with build. Actually, mythi is right, as the sstate-samesigs test belongs to separate pre-stage which is MACHINE independent. Putting it into pre-test which is run in one MACHINE builder script has no meaning, but we kinda agreed we use samesigs test now in concept proving role only. But this also demonstrates that current solution is not flexible: some tests may require separate pre-stage, which we dont provide at all now.

pohly:
@mythi the intention was indeed to treat pre-selftest failures as fatal (after all, something is broken) and then skip the build (thus not wasting time on a change which is broken). This is the same behavior that we also had in Ostro.

Likewise, if the build is broken, there's no need to run post-build-tests.

Regarding machine-specific builds: I think we should aim for one build that covers all machines, via multiconfig. This is more efficient than kicking off n different build jobs which (independently from each other) each have to build, say, some native tools that otherwise could be shared. The downside is we rely more on one big build host, because currently a single build is limited to one build host (albeit some people are using ccache). Another potential downside (or advantage?) is that a build failure on one machine is considered a build failure for all machines. Basically the pressure to keep building working increases a bit, but that's a state that we want to be in anyway.

mythi:
My suggestion are:
- the 'parallel builds' part is dropped. If we multiple MACHINEs are needed later, we'll do that with multiconfig
- for the repo checkout+docker running, we'll have pre selftests stage, build stage, post selftests stage (+what we had earlier)

pohly:
mythi: do you agree to run pre-selftest, build and post-selftest in the same Jenkins job? That is not clear from your comment (because "stages" could be different jobs or different parts of the same job)?

My argument for putting them into one job is that then the selftests can reuse build artifacts from the actual build, without having to rely on sstate sharing. The recipe parser cache also gets reused.

mythi:
On Mon, Mar 13, 2017 at 3:31 PM, Patrick Ohly wrote:

> mythi: do you agree to run pre-selftest, build and post-selftest in the
> same Jenkins job? That is not clear from your comment (because "stages"
> could be different jobs or different parts of the same job)?
>
> My argument for putting them into one job is that then the selftests can
> reuse build artifacts from the actual build, without having to rely on
> sstate sharing. The recipe parser cache also gets reused.
>
>
Yes, same job (I was referring to that with 'for the repo checkout+docker
running')

okartau:
I will remove for() loop of build phase. For now, I will leave MACHINE setting in Jenkinsfile but that can migrate out of CI scripts into config, when/if we decide to go multiconfig.
I will separate oe-selftest calls from build-project script, and they will have separate scripts, called in same job, and in same workspace with build job. Having separate scripts allows to clearly separate which script returns exit status on failure, and also indicate status of steps using clearly named job stages (like those boxes in job page now)

okartau:
test this please

okartau:
test this please

docker/build-project.sh
```diff
@@ -133,14 +137,15 @@ fi
 # Push buildhistory into machine-specific branch in the master buildhistory
 #
 if [ ! -z ${JOB_NAME+x} ]; then
```
mythi: We should move the buildhistory enablement in this block and not keep it enabled in refkit-ci.inc. That'd make the pre/post scripts simpler and closer to what the build is using.

mythi:
@okartau btw, your commits are missing signed-off-by

docker/build-project.sh
okartau: 1. this block runs after bitbake, so what you really mean is to put BH enablement into
block above, near "# Prepare for buildhistory generation" ?

2. should we do these BH enablement changes also in current PR ?

mythi: Yes let's fix that in this PR still.

My point is we don't keep buildhistory settings in refkit-ci.inc anymore. It'd be handled in the same we the archiver mode. The main bitbake build still gets buildhistory if we write the buildhistory settings in this if block.

okartau:
test this please

okartau:
test this please

okartau:
test this please

mythi:
@okartau This looks good to me

@pohly +1 from you too?

PR #55, opened 2017-03-02
meta-iotqa: Disable opencl test for Minnowboard Turbot

skuusela:
The test doesn't work with it and causes this error when running:
drm_intel_gem_bo_context_exec() failed: Device or resource busy
Beignet: "Exec event 0x557ea96c6720 error, type is 4592, error staus is -5"


skuusela:
@ipuustin I wonder how the test has always got passed even though it doesn't work with Minnowboard. Maybe the whole test should be changed?

ipuustin:
The bug is tracked here: https://bugs.freedesktop.org/show_bug.cgi?id=98647

Most likely the example test doesn't return an error value even if the internal kernels fail. Just the values returned are wrong (most likely 0). Best test would be to write a custom OpenCL kernel, execute it, and validate that the result value is what it should be.

PR #58, opened 2017-03-06
Add thermald for thermal management

jlaako:
In order to allow platform thermal management, add thermald to the base
image configuration.

### DIFF
--- a/meta-refkit/classes/refkit-image.bbclass
+++ b/meta-refkit/classes/refkit-image.bbclass
@@ -174,6 +174,10 @@ IMAGE_INSTALL += "util-linux"
 export ALTERNATIVE_PRIORITY_SHADOW ?= "305"
 IMAGE_INSTALL += "shadow"
 
+# Install thermal management daemon by default for
+# handling platform thermal management.
+IMAGE_INSTALL += "thermald"
+
 # Common profile has "packagegroup-common-test" that has packages that are
 # used only in "development" configuration.
 FEATURE_PACKAGES_common-test = "packagegroup-common-test"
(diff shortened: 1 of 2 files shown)


msivosuo:
Please also create Bugzilla item and add an automated test case for thermald.

meta-refkit/classes/refkit-image.bbclass
```diff
@@ -174,6 +174,10 @@ IMAGE_INSTALL += "util-linux"
 export ALTERNATIVE_PRIORITY_SHADOW ?= "305"
 IMAGE_INSTALL += "shadow"
 
+# Install thermal management daemon by default for
+# handling platform thermal management.
+IMAGE_INSTALL += "thermald"
```
mythi: I feel this could go ```MACHINE_EXTRA_RRECOMMENDS/RDEPENDS``` in meta-intel and we'd then install what's in that.

How to deal with the board specific configs? RMC?

jlaako:
Figuring out how to test it needs to wait for next week or so. So far I haven't been trusting user space processes doing thermal management...

meta-refkit/classes/refkit-image.bbclass
jlaako: Are there board specific configs? Why are such needed? I would assume it would work similar way as cpufreq. If you think about normal Linux distros for example there's no way those can know about all possible motherboard/cpu/cooling combinations. Thermal limit points can be read from hardware as well as fan info. So similar thermal profiles can be used as BIOS do (silent, normal, cool).

jlaako: With the boards we currently have hardware does automatic thermal throttling and there are no PWM fans...

mythi: @jlaako Ostro XT has board specific settings for 570x.

mythi:
Let's not add thermald directly in ```IMAGE_INSTALL```.  thermald is part of meta-intel and should therefore be part of ```MACHINE_EXTRA_RDEPENDS/RRECOMMENDS```. 

We'll have to modify our ```IMAGE_INSTALL``` to have packagegroup-machine-base installed instead..

PR #62, opened 2017-03-09
tester-exec.sh: Change hard coded path to variable

skuusela:
Change /home/tester to $HOME variable which makes this script more
portable.


mythi:
@skuusela let's combine #63 and this PR

PR #63, opened 2017-03-09
tester-exec.sh: Minor changes

skuusela:
Report masked tests as 'N/A'. Don't sum masked tests to skipped tests. This way run rate and pass rate of total will be 100% if there is masked tests but no skipped ones. Also change '/home/tester' to $HOME variable which makes this script more portable.


skuusela:
Example:
```
Image: refkit-image-gateway-intel-corei7-64-2017-03-09_03-30-20-build-370.wic
  Device: 570x
  Total:15  Pass:15  Fail:0  Skip:0  Error:0  N/A:5
  Run rate:100%  Pass rate of total:100%  Pass rate of exec:100%
-------------------------------------------------------------------
```

PR #64, opened 2017-03-10
mraa_gpio.py: Add mraa-gpio command output to fail message

skuusela:
Adding the output from the command should help understand why this test
fails when it does.


mythi:
test this please

PR #67, opened 2017-03-14
readme: fix spelling and clarity in README

dbkinder:
Fixed some spelling and grammar issues, and made clarity edits.

Added link to building without docker section.

Clarified only linux build environment supported.

Signed-off-by: David B. Kinder <david.b.kinder@intel.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,21 +2,28 @@ intel-iot-refkit
 ################
 
 This README file contains information on setting up, building, and booting
-Iot Reference OS Kit for Intel(r) Architecture project.
+the IoT Reference OS Kit for Intel(r) Architecture project.
+
+.. note::
+
+   Only Linux development hosts are currently supported, as documented
+   in the `Yocto Project Quick Start Guide`_
+
+.. Yocto Project Quick Start Guide: http://www.yoctoproject.org/docs/current/yocto-project-qs/yocto-project-qs.html#the-linux-distro
 
 Installing Docker on Linux
 ==========================
 
-It's possible to build the project in a Docker container. Using Docker provides
+We recommend you build projects in a Docker container. Using Docker provides
 a packaged and controlled environment for building an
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -2,21 +2,28 @@ intel-iot-refkit
 ################
 
 This README file contains information on setting up, building, and booting
-Iot Reference OS Kit for Intel(r) Architecture project.
+the IoT Reference OS Kit for Intel(r) Architecture project.
+
+.. note::
+
+   Only Linux development hosts are currently support, as documented
```
mythi: supported

README.rst
```diff
@@ -109,7 +119,7 @@ https://git-scm.com/docs/git-submodule
 Installing the Images
 =====================
 
-See detailed instructions in doc/howtos/image-install.rst.
+See detailed instructions in `how to install images <doc/howtos/image-install.rst>`__.
```
gvancuts: Just curious, why do you need two trailing underscore characters?

README.rst
dbkinder: sigh... I shouldn't do these patches when I'm distracted.  Sorry about the spelling error.

README.rst
dbkinder: I was experimenting with "anonymous hyperlinks" (where the reference does not create a lingering name that others could reference too).  In this case it doesn't make a difference so I put it back to the regular single underscore.  While this hyperlinking will work on GitHub, this format of hyperlinking would be a problem if we were using Sphinx, but for now it does give us the ability to link between .rst files in GitHub (normally we'd be linking to the generated .html files and not a .rst file).

gvancuts: Thanks for the explanation @dbkinder . I OK with two trailing underscores, I just did not know what it meant and I now got educated, thanks :-)

mythi:
Merged in #71. Thanks @dbkinder

PR #68, opened 2017-03-15
README.rst: mention possibility to give argument to local build

okartau:
local-build.sh allows to build just subset of images, but we don't
mention this possibility in README.
If started without arguments, target image list is parsed from 
CI settings and will include
all images used in CI, plus test suites.


okartau:
[skip-ci]

PR #73, opened 2017-03-16
imagevariant.bbclass: handle trivial variant/feature name collisions.

klihub:
Don't go into infinite recursion for trivial namespace collision of
variants and IMAGE_FEATUREs. This happens when a variant contains an
IMAGE_FEATURE with an identical name. For instance the following sample
would fail without this patch, recursing on expanding flatpak until
the maximum recursion depth is exceeded:
```
...
IMAGE_FEATURES[validitems] += " \
    flatpak \
    tools-sdk dev-pkgs tools-debug tools-profile \
"

FEATURE_PACKAGES_flatpak = " \
    packagegroup-flatpak \
    ${@'flatpak-predefined-repos' \
         if d.getVar('FLATPAK_APP_REPOS') else ''} \
"
...

IMAGE_VARIANT[flatpak] = "flatpak"
IMAGE_VARIANT[flatpak-sdk] = "flatpak tools-develop tools-debug dev-pkgs"

BBCLASSEXTEND = "imagevariant:flatpak imagevariant:flatpak-sdk"
```
Note that we only handle reasonable to be expected trivial immediate
recursion. Difficult to follow/understand complex cases involving two
or more variant/feature pairs ([foo] = "bar foobar", [bar] = "xyzzy",
[xyzzy] = "foo") we intentionally let fail, although we now give an
error message hinting at the probably cause of the failure.

### DIFF
--- a/meta-refkit/classes/imagevariant.bbclass
+++ b/meta-refkit/classes/imagevariant.bbclass
@@ -19,7 +19,7 @@
 #
 # Example for core-image-minimal.bb:
 #   IMAGE_VARIANT[dev] = "tools-debug tools-profile"
-#   BBCLASSEXTEND = "imagevariant:ptest-pkgs imagevariant:tools-debug,no-debug-tweaks imagevariant=dev"
+#   BBCLASSEXTEND = "imagevariant:ptest-pkgs imagevariant:tools-debug,no-debug-tweaks imagevariant:dev"
 #   ->
 #   core-image-minimal-ptest-pkgs (ptest-pkgs enabled)
 #   core-image-minimal-tools-debug-no-debug-tweaks (debug-tweaks disabled)
--- a/meta-refkit/classes/imagevariant.bbclass
+++ b/meta-refkit/classes/imagevariant.bbclass
@@ -46,13 +46,30 @@ python imagevariant_virtclass_handler () {
     # Expand parameter aliases, recursively.
     def expand(parameters):
         modified = []
-        for parameter in parameters:
+        recursive = []
+
+        for parameter in parameters or []:
             alias = e.data.getVarFlag('IMAGE_VARIANT', parameter, True)
+
             if alias is not None:
-                modified.extend(expand(alias.split()))
+                split_alias = alias.split()
+                if parameter in split_alias:
+                    # continue expansion with parameter removed to avoid
+                    # infinite recursion
+                    split_alias.remove(parameter)
+                    recursive.append(parameter)
+                try:
+                    modified.extend(expand(split_alias))
+                except RuntimeError as re:
+                    bb.fatal('You probably have a non-trivially ' \
+                             'expandable imagevariant (%s). ' \
+                             'Runtime error: %s.' % (variant, re))
             else:
                 modified.append(parameter)
+
+        modified.extend(recursive)
         return modified
+
     parameters = expand(parameters)
 
     # Validate parameters and apply them to IMAGE_FEATURES.


pohly:
Can you give an example where the current code fails in the commit message? That would make the problem more obvious.

In the code, can you add a comment about the "if parameter in split_alias" (i.e. when it is needed)? I think I understand what it does, but it's not obvious.

When you say, "we intentionally do let it fail", what is the error that one gets? It might be worth mentioning this limitation in the description of the IMAGE_FEATURE variable.

klihub:
Sure, I'll update.

klihub:
The 'intentionally let fail' part is implemented by not being specially handled. So it will fail with an maximum recursion depth exceeded error with file/line information for the recursive expand call.

pohly:
Can you add a comment to the code that shows up in the infinite recursion case that explains the root cause? That'll save the next person who accidentally runs into this some head-scratching.

klihub:
BTW, it'd be nicer to not bb.fatal immediately out, just bb.error first, then finally bb.fatal out once all recursion errors have been reported. This way, all the participants in non-trivial recursion chains would be reported instead of just the first detected one.

I have a slightly modified patch that does that using event handlers... but since the code uses e.data.setVar for error bookkeeping and I'm afraid it is too ugly to be acceptable. Plus I could immediately figure out how to make it report a single fatal instead of as many as there are recursion errors.

It'd be something like this:
```
...
               try:
                    modified.extend(expand(split_alias))
                except RuntimeError as re:
                    bb.error('You probably have a non-trivially ' \
                             'recursive imagevariant (%s). ' \
                             'Runtime error: %s.' % (variant, re))
                    e.data.setVar('IMAGEVARIANT_ERRORS', 'true')
...
python imagevariant_check_errors () {
    if e.data.getVar('IMAGEVARIANT_ERRORS'):
        bb.fatal('Exiting due to imagevariant errors.')
}

addhandler imagevariant_check_errors
imagevariant_check_errors[eventmask] = "bb.event.RecipeParsed"
```

If someone more bitbake-initiated can suggest a better working way to accomplish the same idea, I'm all ears.

pohly:
Instead of detecting just one level of recursion by checking the current parameter against its own replacement, wouldn't it be fairly easy to add a parameter to expand() which tracks all the already encountered parameters? Pass that through into all recursive expand() calls and recursion can be detected at all levels. Then there's no need to improve error reporting (a single message, as you said, is complicated and probably not easily doable).

mythi:
@klihub sorry for catching this up late. I'm OK with the change. However, I'd think 'non-trivially expandable'   (or something like that) would be more descriptive error note.

It looks all other review feedback is address in d9342a9

klihub:
I'll update the error message to the better one suggested by @mythi ...

@pohly I'm not sure if that's really a good idea... I did not want to allow arbitrarily obfuscated cross-recursive variant/feature chains. Just ones that are reasonable to expect/natural to have. The basic problem is that imagevariant.bbclass does not provide a way, or even consciously try, to differentiate between features and variants... Yet it tries to allow variant definitions to reference other variant names in addition to features. Hence the recursion and the ambiguity caused by the namespace clash of features and variants. This is exactly why it fell into an infinite recursion in the first place. There are two ways to resolve the ambiguity: 1) provide a notation for explicitly stating which namespace one is referring to, or 2) to do automatic disambiguation when it is obvious and trivial. I chose the latter because it required minimal changes and it did not allow unreasonably obfuscated definitions.

PR #75, opened 2017-03-16
[TEST BUILD] Fix master build "ci_git_commit" file not found

skuusela:
Create the file in the same context as the reading the file.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -123,7 +123,8 @@ try {
                 tester_script = readFile "docker/tester-exec.sh"
                 testinfo_data["${target_machine}"] = readFile "${target_machine}.testinfo.csv"
 
-                if ( !is_pr ) {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -123,7 +123,8 @@ try {
                 tester_script = readFile "docker/tester-exec.sh"
                 testinfo_data["${target_machine}"] = readFile "${target_machine}.testinfo.csv"
 
-                if ( !is_pr ) {
+                if ( is_pr ) {
```
mythi: did the logic change intentionally?

skuusela: Yes, for testing.

okartau:
CI_GIT_COMMIT is also used for another thing, in BUILDHISTORY branch name so we should not remove creation of this variable in build script

PR #76, opened 2017-03-16
CI build-project.sh: fix location of ci_git_commit file

okartau:
Recent change moved creation of ci_git_commit into later block
which runs after env-init, so that file was created one directory
level deeper, in build/ directory. But as other parts look for
file at top level of WORKSPACE, thise broke master build
where this file is used for summary message.
Code is safer if we spefify full path in writing.


okartau:
test this please

PR #78, opened 2017-03-17
refkit: use static DISTRO_VERSION

mythi:
Currently, our DISTRO_VERSION is set to a misleading value and is not even
updated regularly. As we are working in a (partially) rolling-release mode,
any "point release" versioning does not make much sense.

Use static DISTRO_VERSION that gives no indication of any specific version.

### DIFF
--- a/meta-refkit/classes/refkit-image.bbclass
+++ b/meta-refkit/classes/refkit-image.bbclass
@@ -384,12 +384,10 @@ ROOTFS_POSTPROCESS_COMMAND += "${@'extra_motd;' if d.getVar('REFKIT_EXTRA_MOTD',
 # FIXME: make this work with both ${IMAGE_ROOTFS}/etc/ and ${IMAGE_ROOTFS}/usr/lib
 refkit_image_patch_os_release () {
     sed -i \
-        -e 's/distro-version-to-be-added-during-image-creation/${DISTRO_VERSION}/' \
         -e 's/build-id-to-be-added-during-image-creation/${BUILD_ID}/' \
         ${IMAGE_ROOTFS}/etc/os-release
 }
 refkit_image_patch_os_release[vardepsexclude] = " \
-    DISTRO_VERSION \
     BUILD_ID \
 "
 ROOTFS_POSTPROCESS_COMMAND += "refkit_image_patch_os_release; "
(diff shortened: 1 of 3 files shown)


gvancuts:
Hi @mythi , do I understand this correctly that there would still be a `BUILD_ID` in `/etc/os-release` that would allow anyone to easily identify what build version is currently running?

mythi:
@gvancuts correct. For local non-Docker builds that'll be ```${DATE}```.

okartau:
isnt it yocto variable ${DATETIME} for local builds ?
Also, I think Docker does (should) not make difference, it's CI.

In non-CI local build /etc/os-release will look like this (with current PR changes applied,
just tried on a local system, non-docker build):
ID="refkit"
NAME="IoT Reference OS Kit for Intel(r) Architecture"
VERSION="refkit.0"
VERSION_ID="refkit.0"
PRETTY_NAME="IoT Reference OS Kit for Intel(r) Architecture refkit.0"
BUILD_ID="20170317100231"

In a CI build BUILD_ID gets value from Jenkins-generated CI_BUILD_ID-build-BUILD_NR
I think other values will be same as local-build (?)

okartau:
I peeked into PR verification build image to get contents of /etc/os-release in CI-build (uses Docker) case,
and as expected other values are same as above, with BUILD_ID being different:
BUILD_ID="2017-03-17_00-27-18-build-448"

mythi:
@pohly could you do a peer review for this?

okartau:
I got myself messed into overloaded  meaning of "Iocal build".
We have 2 kinds of builds that can be called "local build":
1. build started with local-build.sh
This forces Docker use, and sets some values including BUILD_ID
maintaining (incrementing) local build-number locally in dot-file
result will be similar to what CI build has: 
BUILD_ID="2017-03-17_14-51-59-build-7"

2. building without Docker, using direct bitbake call (this method also documented in README)
BUILD_ID value will not be pre-set, resulting BUILD_ID taking 
value from Yocto DATETIME
BUILD_ID="20170317100231"

For completeness, 3rd kind of build is CI build, in Docker,
with Jenkins-managed up-counting BUILD_NUMBER:
BUILD_ID=2017-03-17_04-58-09-build-146

Is there a issue that builds of type 1. and 2. 
get different-looking BUILD_IDs ?

mythi:
test this please

PR #79, opened 2017-03-17
remove swupd tweaks from default build

pohly:
We don't even have meta-swupd in refkit, so it makes more sense to
leave distro settings at their defaults instead of enabling
swupd-related tweaks in all cases.


mythi:
@pohly Thanks!

PR #80, opened 2017-03-17
rmc: enable serial console when running under QEMU

rojkov:
Since we can run ordinary intel-corei7-64 wic images under QEMU
it makes sense to add QEMU quirks enabling serial console
to the RMC DB.

This patch appends "console=ttyS0,115200" to kernel command
line in case the image is used to run a QEMU VM e.g. with the
command:

    $ runqemu ovmf refkit-image-common wic intel-corei7-64 slirp serial nographic

Unfortunately different major versions of QEMU produce different
RMC fingerprints, so the QEMU versions 2.6.x and 2.8.x should be
treated as different boards.


pohly:
With this change in place, how many getty serial instances are active? There have been problems in the past with getty@ttyS1/S2/S3 (?) failing and respawning, which is annoying.

The other, conceptual problem is that qemu gets invoked with a varying number of serial consoles active. I'm pretty sure (but haven't looked recently) that oe-selftest runs qemu with two serial consoles and waits for a console prompt on the second serial console. We can still fix that if it's a problem later on.

rojkov:
Like I commented in https://bugzilla.yoctoproject.org/show_bug.cgi?id=11061 meta-intel wants to spawn getty for ttyS0,ttyS1 and ttyS2 for `intel-corei7-64` machines. It doesn't know about QEMU having only two serial ports. POSTINSTALL.sh is supposed to be used for dropping unneeded getty's, but it's ignored in the current Refkit setup IIUC.

rojkov:
Again IIUC oe-selftest shouldn't have problems with detecting a console prompt because getty instances are started successfully for ttyS0 and ttyS1 under QEMU. `console=` in kernel cmdline is not relevant here.

pohly:
According to "https://wiki.archlinux.org/index.php/Systemd_FAQ", systemd doesn't use /etc/inittab, which is what the POSTINSTALL.sh from http://meta-intel.yoctoproject.narkive.com/dGV67zHQ/patch-1-1-rmc-add-installer-config-and-postinstall-sh-for-broxton-m-boards modifies.

This POSTINSTALL.sh looks like a gross hack to me which only works under very special circumstances. Let's not bother trying to support it.

We probably end up with all gettys enabled because SERIAL_CONSOLES is checked while building a systemd-enabled image. That's good enough for now until someone can think of a better solution.

mythi:
@rojkov this looks reasonable to me but we could still check whether meta-intel would accept the qemu fingerprints too. btw, I've filed an enhancement in bugzilla that RMC is able to manage multiple fingerprints per board: https://bugzilla.yoctoproject.org/show_bug.cgi?id=11133. Seems like that is valid for qemu too.

I'll add more comments in https://bugzilla.yoctoproject.org/show_bug.cgi?id=11061 so let's move the discussion there.

rojkov:
@mythi Also submitted to meta-intel (https://lists.yoctoproject.org/pipermail/meta-intel/2017-March/004675.html).

mythi:
We're going to get this from meta-intel.

rojkov:
They haven't merged the patch to meta-intel yet.

mythi:
@rojkov it's queued for merging.

PR #81, opened 2017-03-17
CI tester-exec.sh: streamline and reduce unused legacy code

okartau:
Variables were initialized one-by-one to avoid error
on use as -u is used in shell. This consumes unneccessary code
lines but also creates meaningless job log files, so combine
multiple vars into declare stmt. Also, utilize += in adding.

Unused legacy code lines about older devices can be removed


okartau:
test this please

PR #88, opened 2017-03-23
refkit images: mount read/write by default

pohly:
Mounting as read-only was enforced with the argument that then
fsck can be applied to the rootfs, but exactly that was explicitly
disabled in the distro configuration.

Disabling fsck makes sense, but that is better set in the image class,
and then we can also directly mount rw (avoids a remount, better for a
stateless image, for example because we don't need the empty
/etc/machine-id).


mythi:
test this please

mythi:
test this please

mythi:
test this please

PR #92, opened 2017-03-27
Reduce Python 3 dependencies

ipuustin:
Compile Python 3 without readline, gdbm, and db support. This removes the need for GPLv3 depedencies.

The patch was submitted upstream (https://patchwork.openembedded.org/patch/136368/) but never merged. This is the `.bbappend` version of it.

### DIFF
--- a/meta-refkit/conf/distro/refkit.conf
+++ b/meta-refkit/conf/distro/refkit.conf
@@ -71,6 +71,9 @@ PACKAGECONFIG_remove_pn-gstreamer1.0-plugins-base = "pango"
 
 PACKAGECONFIG_remove_pn-pulseaudio = "avahi"
 
+# leave out readline, gdbm, and db
+PACKAGECONFIG_pn-python3 = ""
+
 # since we have OpenCL support, turn it on
 PACKAGECONFIG_append_pn-opencv = " opencl"
 PREFERRED_PROVIDER_virtual/opencl-headers = "opencl-headers"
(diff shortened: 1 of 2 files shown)


mythi:
@ipuustin I think this is OK but perhaps follow-up with Ross if it was simply forgotten.

meta-refkit/conf/distro/refkit.conf
```diff
@@ -62,6 +62,9 @@ PACKAGECONFIG_remove_pn-gstreamer1.0-plugins-base = "pango"
 
 PACKAGECONFIG_remove_pn-pulseaudio = "avahi"
 
+# leave out readline, gdbm, and db
+PACKAGECONFIG_pn-python3 = ""
```
mythi: ```?=``` is preferred.

ipuustin: Right.

ipuustin:
@rossburton: what's your take on this? Should we have this in refkit or could we add the functionality to oe-core?

rossburton:
Sorry, slipped through.  Yes, I think the PACKAGECONFIGs should be in core.  It's very late but I'll review that properly and see what we can do.

mythi:
This is on it's way to oe-core but let's take the bbappend in until then.

ipuustin:
It seems that this is somewhat broken now due to https://github.com/openembedded/openembedded-core/commit/54a2549802a911cad2475a6aa379315a834419d8 . Need to investigate a bit.

PR #94, opened 2017-03-28
Test QEMU

skuusela:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -92,6 +92,7 @@ try {
                     }
                 } // docker_image
                 tester_script = readFile "docker/tester-exec.sh"
+                qemu_script = readFile "docker/run-qemu.exp"
                 testinfo_data["${target_machine}"] = readFile "${target_machine}.testinfo.csv"
                 if ( !is_pr ) {
                     ci_git_commit = readFile("ci_git_commit").trim()
@@ -119,6 +120,7 @@ try {
                     deleteDir() // clean workspace
                     echo "Testing test_${test_device} with image_info: ${one_target_testinfo}"
                     writeFile file: 'tester-exec.sh', text: tester_script
+                    writeFile file: 'run-qemu.exp', text: qemu_script
                     // append newline so that tester-exec.sh can parse it using "read"
                     one_target_testinfo += "\n"
                     // create testinfo.csv on this tester describing one image
@@ -127,7 +129,7 @@ try {
                         withEnv(["CI_BUILD_ID=${ci_build_id}",
                             "MACHINE=${test_machine}",
                             "TEST_DEVICE=${test_device}" ]) {
-                                sh 'chmod a+x tester-exec.sh && ./tester-exec.sh'
+                                sh 'chmod a+x tester-exec.sh run-qemu.exp && ./tester-exec.sh'
                         }
                     } catch (Exception e) {
                         throw e
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,18 +16,6 @@
 
 def is_pr = env.JOB_NAME.endsWith("_pull-requests")
 def target_machine = "intel-corei7-64"
-def test_devices = [ "570x", "minnowboardturbot" ]
-
-// mapping [HW_target:build_MACHINE]
-// for tester job to select which MACHINE img to test.
-// note: I'd like to define just one map structure
-// combining info that is now in test_devices[]  and mapping[],
-// but pipeline seems to have no support yet
-// for iterating cleanly over map keys.
-def mapping = [
-  '570x' : 'intel-corei7-64',
-  'minnowboardturbot' : 'intel-corei7-64'
-]
 
 // JOB_NAME expected to be in form <layer>_<branch>
 def current_project = "${env.JOB_NAME}".tokenize("_")[0]
@@ -117,52 +105,46 @@ try {
 
     // find out combined size of all testinfo files
     testinfo_sumz += testinfo_data["${target_machine}"].length()
+    test_targets = testinfo_data["${target_machine}"].split("\n")
     // skip tester parts if no tests configured
     if ( testinfo_sumz > 0 ) {
         def test_runs = [:]
-        for(int i = 0; i < test_devices.size(); i++) {
-            def test_device = test_devices[i]
-            // only if built for machine that this tester wants
-            if ( target_machine == mapping["${test_device}"] ) {
-                // testinfo_data may contain multiple lines stating different images
-                String[] separated_testinfo = testinfo_data["${target_machine}"].split("\n")
-                for (int m = 0; m < separated_testinfo.length; m++) {
-                    def one_image_testinfo = separated_testinfo[m]
-                    echo "Image #${m} to be tested on test_${test_device} info: ${separated_testinfo[m]}"
-                    test_runs["test_${m}_${test_device}"] = {
-                        node('refkit-tester') {
-                            deleteDir() // clean workspace
-                            echo "Testing test_${test_device} with image_info: ${one_image_testinfo}"
-                            writeFile file: 'tester-exec.sh', text: tester_script
-                            // append newline so that tester-exec.sh can parse it using "read"
-                            one_image_testinfo += "\n"
-                            // create testinfo.csv on this tester describing one image
-                            writeFile file: "testinfo.csv", text: one_image_testinfo
-                            def img = one_image_testinfo.split(",")[0]
-                            try {
-                                withEnv(["CI_BUILD_ID=${ci_build_id}",
-                                    "MACHINE=${mapping["${test_device}"]}",
-                                    "TEST_DEVICE=${test_device}" ]) {
-                                        sh 'chmod a+x tester-exec.sh && ./tester-exec.sh'
-                                }
-                            } catch (Exception e) {
-                                throw e
-                            } finally {
-                                // read tests summary prepared by tester-exec.sh
-                                // Here one tester adds it's summary piece to the global buffer.
-                                global_sum_log += readFile "results-summary-${test_device}.${img}.log"
-                                archiveArtifacts allowEmptyArchive: true,
-                                                 artifacts: '*.log, *.xml'
-                            }
-                            // without locking we may lose tester result set(s)
-                            // if testers run xunit step in nearly same time
-                            lock(resource: "step-xunit") {
-                                step_xunit()
-                            }
-                        } // node
-                    } // test_runs =
-                } // for
-            } // if target_machine == mapping
+        for(int i = 0; i < test_targets.size(); i++) {
+            def one_target_testinfo = test_targets[i]
+            def test_device = one_target_testinfo.split(',')[4]
+            def test_machine = one_target_testinfo.split(',')[3]
+            def img = one_target_testinfo.split(",")[0]
+            test_runs["test_${i}_${test_device}"] = {
+                node('refkit-tester') {
+                    deleteDir() // clean workspace
+                    echo "Testing test_${test_device} with image_info: ${one_target_testinfo}"
+                    writeFile file: 'tester-exec.sh', text: tester_script
+                    // append newline so that tester-exec.sh can parse it using "read"
+                    one_target_testinfo += "\n"
+                    // create testinfo.csv on this tester describing one image
+                    writeFile file: "testinfo.csv", text: one_target_testinfo
+                    try {
+                        withEnv(["CI_BUILD_ID=${ci_build_id}",
+                            "MACHINE=${test_machine}",
+                            "TEST_DEVICE=${test_device}" ]) {
+                                sh 'chmod a+x tester-exec.sh && ./tester-exec.sh'
+                        }
+                    } catch (Exception e) {
+                        throw e
+                    } finally {
+                        // read tests summary prepared by tester-exec.sh
+                        // Here one tester adds it's summary piece to the global buffer.
+                        global_sum_log += readFile "results-summary-${test_device}.${img}.log"
+                        archiveArtifacts allowEmptyArchive: true,
+                                         artifacts: '*.log, *.xml'
+                    }
+                    // without locking we may lose tester result set(s)
+                    // if testers run xunit step in nearly same time
+                    lock(resource: "step-xunit") {
+                        step_xunit()
+                    }
+                } // node
+            } // test_runs =
         } // for i
         stage('Parallel test run') {
             set_gh_status_pending(is_pr, 'Testing')
--- a/docker/tester-exec.sh
+++ b/docker/tester-exec.sh
@@ -117,7 +117,7 @@ TEST_SUITE_FOLDER_URL=${CI_BUILD_URL}/testsuite/${MACHINE}
 
 # get necessary params from testinfo.csv file written to tester workspace
 # by code in Jenkinsfile. We have just one line for this tester session.
-while IFS=, read _img _tsuite _tdata _mach
+while IFS=, read _img _tsuite _tdata _mach _dev
 do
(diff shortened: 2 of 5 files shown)


skuusela:
test this please

skuusela:
test this please

skuusela:
test this please

skuusela:
test this please

skuusela:
test this please

Jenkinsfile
```diff
@@ -86,6 +86,41 @@ try {
                         stage('Build') {
                             sh "${params}"
                         }
+                        
+                        set_gh_status_pending(is_pr, 'Testing with QEMU')
+                        params = ["${script_env_global}", "${script_env_local}",
+                                  "docker/test-qemu.sh"].join("\n")
+                        stage("Test QEMU") {
```
okartau: I think testing should run after "store images" stage, because testing can change state of build results directory

docker/test-qemu.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
```
okartau: python script should not be named .sh

Jenkinsfile
```diff
@@ -86,6 +86,41 @@ try {
...
+                            try{
+                                sh "${params}"
+                            } catch (Exception e) {
+                                throw e
+                            } finally{
+                                archiveArtifacts allowEmptyArchive: true,
```
okartau: should we rename TEST-* files from QEMU testing similar to renaming after HW-based testing? Otherwise QEMU test results files will be named just TEST-oeqa.runtime-* and it remains unclear on which were these produced.

mythi:
meta-intel has http://git.yoctoproject.org/cgit/cgit.cgi/meta-intel/commit/?id=2a955487d8563b929846da01a69ed6d0f560bd5b so c689faa is not needed. For testing purposes I'd suggest you drop c689faa and add meta-intel submodule update instead.

docker/opensuse-42.2/Dockerfile
```diff
@@ -10,7 +10,8 @@ RUN zypper --no-color --non-interactive ref -f && \
     zypper --no-color --non-interactive install --no-recommends python gcc \
             gcc-c++ git chrpath make wget python-xml diffstat makeinfo \
             python-curses patch socat libSDL-devel tar which unzip \
-            python3 python3-curses glibc-locale syslinux
+            python3 python3-curses glibc-locale syslinux python3-devel \
+            python3-six python3-netifaces python3-unittest-xml-reporting
```
mythi: The commit message confuses me. What are the tests needing these and why these are not needed already now?

skuusela: The QEMU tests are run inside docker and real hardware tests are run on different server natively.

mythi:
I'd think we should evaluate the following two options to do qemu testing:

1. Qemu as a test target in ```def test_devices = [ "570x", "minnowboardturbot" ]```
2. Qemu via ```-c testimage``` with our test-suites in build-project.sh/post-build.sh

skuusela:
test this please

skuusela:
@mythi Thoughts about this way of testing QEMU? This way QEMU is just one addition to `test_devices[...]` and short and simple testing script for QEMU is added to tester-exec.sh. Things I'm not sure about are the addition of debug-tweaks to REFKIT_IMAGE_EXTRA_FEATURES and also the way ports are chosen for QEMU so they won't collide with different test runs.

meta-refkit/conf/distro/include/refkit-development.inc
```diff
@@ -4,7 +4,7 @@ IMA_EVM_KEY_DIR = "${IMA_EVM_BASE}/data/debug-keys"
 # Enable local root access. Always use _append, to
 # avoid getting this change overwritten by a
 # REFKIT_IMAGE_EXTRA_FEATURES = "foo" in local.conf.
-REFKIT_IMAGE_EXTRA_FEATURES_append = " autologin"
+REFKIT_IMAGE_EXTRA_FEATURES_append = " autologin debug-tweaks"
```
pohly: Why is this necessary for testing under qemu? Can't the approach used for other target hardware also be used here?

If that doesn't work (for example, because the image never gets installed for qemu), then how about "installing" a .ssh/authorized_keys via serial console? We have "autologin" active, so the serial console should be usable.

skuusela: With real hardware we first flash the image to emmc/sd card and then add the ssh-key to authorized_keys but it's done by DAFT program that isn't used with QEMU. So I thought this might be the simplest solution but I understand why it couldn't be used. So I guess I'll try to add ssh-key with serial console.

mythi:
@skuusela This approach looks good to me. I agree with @pohly on ```debug-tweaks```.  Why is 3161014 necessary? I can see you wanted to 'test with qemu on certain images only' but is there a reason (other than the .mask list might get too long).

docker/tester-exec.sh
```diff
@@ -65,10 +65,34 @@ testimg() {
   grep -Fvxf ${MASKFILE} ${MANIFEST} > tmp && mv tmp ${MANIFEST}
 
   # Execute with +e to make sure that possibly created log files get
-  # renamed, archived, published even when AFT or some of renaming fails
+  # renamed, archived, published even when DAFT, QEMU or some of renaming fails
   set +e
-  daft ${DEVICE} ${FILENAME} --record
-  AFT_EXIT_CODE=$?
+  if [ "${DEVICE}" = "qemu" ]; then
+    wget ${_WGET_OPTS} ${CI_BUILD_URL}/images/${MACHINE}/ovmf.qcow2
```
mythi: to make it easier to read, this block could be a function.

skuusela:
@mythi I thought we only wanted to test common profile with qemu? That was the reasoning with 3161014.

docker/tester-exec.sh
```diff
@@ -65,10 +65,34 @@ testimg() {
...
+
+    # Make port numbers that won't collide with other test runs
+    PORT1=$((21000+20*${BUILD_NUMBER: -1}+$BUILD_NUMBER+$TEST_ID))
+    PORT2=$((25000+20*${BUILD_NUMBER: -1}+$BUILD_NUMBER+$TEST_ID))
+
+    qemu-system-x86_64 -device virtio-net-pci,netdev=net0,mac=52:54:00:12:35:02 \
```
mythi: worth enabling kvm

docker/tester-exec.sh
```diff
@@ -15,6 +15,31 @@
...
 
+# function to test and image with QEMU
+test_qemu() {
+  wget ${_WGET_OPTS} ${CI_BUILD_URL}/images/${MACHINE}/ovmf.qcow2
+
+  # Make port numbers that won't collide with other test runs
+  PORT1=$((21000+20*${BUILD_NUMBER: -1}+$BUILD_NUMBER+$TEST_ID))
+  PORT2=$((25000+20*${BUILD_NUMBER: -1}+$BUILD_NUMBER+$TEST_ID))
+
+  qemu-system-x86_64 -device virtio-net-pci,netdev=net0,mac=52:54:00:12:35:02 \
```
okartau: should consider putting some hashed values into MAC address as well?
I think that in default config, net device will not be isolated, (could be isolated with restrict option)
so any concurrent qemu sessions would use colliding MAC address which may lead to trouble.

skuusela:
@mythi Could I just add ssh keys to this repo and add them to development images at build stage? Just like with this [private key](https://github.com/intel/intel-iot-refkit/blob/master/meta-refkit/files/dm-verity/private.pem) used in [refkit-development.inc](https://github.com/intel/intel-iot-refkit/blob/master/meta-refkit/conf/distro/include/refkit-development.inc). I think that approach is cleaner than adding the keys to the image later on.

mythi:
@skuusela That approach would take us to same as ```debug-tweaks``` basically. Also, the key would be different from what the testing harness uses.

I might be taking a side-track here but I was wondering can we somehow move qemu run behind DAFT and use the 'PC host' + pcdevice testing harness? Then all the testing harness code can be used. Just a thought.

docker/tester-exec.sh
```diff
@@ -15,6 +15,31 @@
...
+  unique_port=$((pid%5000)) # Make a number between 0-4999
+  unique_mac=$((100000+(pid%900000))) # Make a number with 6 digits
+  PORT1=$((10000+unique_port))
+  PORT2=$((15000+unique_port))
+  QEMU_MAC=52:54:00:${unique_mac:0:2}:${unique_mac:2:2}:${unique_mac:4:2}
+
+  ./run-qemu.sh $PORT1 $PORT2 $QEMU_MAC $FILENAME &
+
+  QEMU_PID=$!
+  sleep 20 # Wait some time for QEMU to boot
```
mythi: the sleep should be unnecessary because run-qemu.sh guarantees it's up

skuusela: @mythi QEMU might be up but the ssh key might not have been added yet and if the tests are started some of them may fail.

mythi: @skuusela there's another sleep in run-qemu.sh that looks like it's for the same reason. Are they both necessary?

skuusela: @mythi The reason for that sleep is that otherwise the expect script and the qemu will exit and the tests can't be run. There might be better solution but for now it will do.

skuusela:
Testing for stability, test this please

skuusela:
test this please

skuusela:
test this please

docker/tester-exec.sh
```diff
@@ -27,14 +29,10 @@ test_qemu() {
...
-  qemu-system-x86_64 -device virtio-net-pci,netdev=net0,mac=$QEMU_MAC \
-  -netdev user,id=net0,hostfwd=tcp::$PORT1-:22,hostfwd=tcp::$PORT2-:23 \
-  -drive if=virtio,file=$FILENAME,format=raw -no-reboot -vga vmware \
-  -device virtio-rng-pci -drive if=pflash,format=qcow2,file=ovmf.qcow2 \
-  -vga std -cpu kvm64 -enable-kvm -m 256 -serial null --nographic &
+  nohup ./run-qemu.sh $PORT1 $PORT2 $QEMU_MAC $FILENAME &>run_qemu.log &
 
   QEMU_PID=$!
-  sleep 10 # Wait some time for QEMU to boot
+  sleep 20 # Wait some time for QEMU to boot and add an ssh key to image
```
mythi: could this be replaced with a test ssh attempt using a long timeout ```ssh -o ConnectTimeout 100```

skuusela: @mythi The connection seems to get refused if ssh connection is tried right away and if there is a small sleep before ssh, there will be password prompt. So I'll need to make a while loop if that is really needed. But first I'll try to make this stable. What I have tested that sleep doesn't affect it.

skuusela:
test this please

skuusela:
test this please

skuusela:
test this please

skuusela:
test this please

docker/run-qemu.sh
```diff
@@ -0,0 +1,20 @@
...
+
+spawn qemu-system-x86_64 -device virtio-net-pci,netdev=net0,mac=$QEMU_MAC \
+-netdev user,id=net0,hostfwd=tcp::$PORT1-:22,hostfwd=tcp::$PORT2-:23 \
+-drive if=virtio,file=$FILENAME,format=raw -no-reboot -vga vmware \
+-device virtio-rng-pci -drive if=pflash,format=qcow2,file=ovmf.qcow2 \
+-vga std -cpu kvm64 -enable-kvm -m 256 -serial mon:stdio --nographic
+
+expect "root@intel-corei7-64:~#"
+# Wait for boot to finish completely
+sleep 5
```
skuusela: Hopefully this sleep was the thing that finally got this stable as previously most of the time the QEMU testing failed. I still want to test this in CI few times more to make sure. test this please

mythi: @skuusela do you have the logs to understand why it's failing?

skuusela:
test this please

docker/run-qemu.sh
```diff
@@ -0,0 +1,18 @@
...
+set FILENAME [lindex $argv 3]
+
+spawn qemu-system-x86_64 -device virtio-net-pci,netdev=net0,mac=$QEMU_MAC \
+-netdev user,id=net0,hostfwd=tcp::$PORT1-:22,hostfwd=tcp::$PORT2-:23 \
+-drive if=virtio,file=$FILENAME,format=raw -no-reboot -vga vmware \
+-device virtio-rng-pci -drive if=pflash,format=qcow2,file=ovmf.qcow2 \
+-vga std -cpu kvm64 -enable-kvm -m 256 -serial mon:stdio --nographic
+
+set timeout 60
+expect "root@intel-corei7-64:~#"
```
skuusela: @mythi Ok I finally found the real reason why it didn't work previously. On default expect uses 10 seconds timeout for the `expect` command. So sometimes the booting took more than 10 seconds and it skipped the `expect "root@intel-corei7-64:~#"` and right away `send "echo -e '$SSH_KEY' > /home/root/.ssh/authorized_keys\n"`. And obviously that command didn't work as QEMU hadn't gone past the login yet.

skuusela:
@mythi I tried testing all the profile images with QEMU and all the tests passed (all that I didn't mask). I also tested that if there is a problem with QEMU and testing can't be done the script exits nicely. So this PR is now working as it should. Any changes still needed?

mythi:
@skuusela great work! The remaining feedback I have is is related to the intermediate changes in these three commits.  I'd think squashing the commits in two would be preferred: 4417335 without the qemu parts and then qemu testing is added in one.

meta-refkit/conf/distro/include/refkit-ci.inc
```diff
@@ -74,13 +74,17 @@ REFKIT_CI_ESDK_TARGETS=""
...
 # Space separated list of tuples, each should in format:
-# <image_name>,<testsuite_name>,<testfiles_name>,$MACHINE
+# <image_name>,<testsuite_name>,<testfiles_name>,$MACHINE,<test_device_name>
 REFKIT_CI_TEST_RUNS=" \
-  refkit-image-common,iot-testsuite.tar.gz,iot-testfiles.${MACHINE}.tar.gz,${MACHINE} \
-  refkit-image-computervision,iot-testsuite.tar.gz,iot-testfiles.${MACHINE}.tar.gz,${MACHINE} \
-  refkit-image-gateway,iot-testsuite.tar.gz,iot-testfiles.${MACHINE}.tar.gz,${MACHINE} \
+  refkit-image-common,iot-testsuite.tar.gz,iot-testfiles.${MACHINE}.tar.gz,${MACHINE},570x \
+  refkit-image-common,iot-testsuite.tar.gz,iot-testfiles.${MACHINE}.tar.gz,${MACHINE},minnowboardturbot \
+  refkit-image-common,iot-testsuite.tar.gz,iot-testfiles.${MACHINE}.tar.gz,${MACHINE},qemu \
```
mythi: minor feedback: qemu here should be in the other commit still

skuusela: @mythi Oh yes, thanks.

mythi:
@skuusela one last question :) did you try the test report email works OK?

skuusela:
@mythi Oh yeah, I'll try that before this gets merged.

skuusela:
The email generation seems to work as intended (N/A tests are the ones QEMU can't run from qemu.mask):
```
Test results:

Image: refkit-image-common-intel-corei7-64-2017-04-26_06-00-31-build-833.wic
  Device: qemu
  Total:15  Pass:15  Fail:0  Skip:0  Error:0  N/A:4
  Run rate:100%  Pass rate of total:100%  Pass rate of exec:100%
```

PR #96, opened 2017-03-29
Always send master build email

skuusela:
Currently if there is an error in testing the email won't get sent. Move
the code block that sends the email under 'finally' so it always gets sent.
Also if testing has an error build result will stay as success.
Manually set it as 'FAILURE' if there is an error at any point of the
build and set it as 'UNSTABLE' if there is an error in testing (some problem with testing setup itself.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -188,29 +188,18 @@ try {
         }
     } // if testinfo_sumz
 
-    echo "After test stage: build result is ${currentBuild.result}"
-
 } catch (Exception e) {
     echo "Error: ${e}"
-    if (is_pr) {
-        // GH API cant take more than 140 chars in status msg so lets truncate.
-        def _msg = e.getMessage().take(120)
-        setGitHubPullRequestStatus state: 'FAILURE', context: "${env.JOB_NAME}", message: "Build failed: ${_msg}"
-    }
     throw e
 } finally {
+    //  If tests stage was skipped because of no tests, then currentBuild.result
+    //  remains null until end so manually set it as SUCCESS
+    if (currentBuild.result == null) {
+        currentBuild.result == 'SUCCESS'
+    }
     echo "Finally: build result is ${currentBuild.result}"
     if (is_pr) {
-        // need to cross-check build result to handle possible combinations:
-        // 1. FAILURE in xUnit processing does not cause Exception block below
-        //   to be run, but currentBuild.result is correctly set to FAILURE.
-        // 2. If tests stage was skipped because of no tests,
-        //   then currentBuild.result remains null until end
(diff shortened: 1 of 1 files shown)


skuusela:
test this please

skuusela:
test this please

mythi:
@skuusela is this ready?

skuusela:
@mythi Yes it's been tested and ready for merge.

Jenkinsfile
```diff
@@ -192,6 +192,8 @@ try {
 
 } catch (Exception e) {
     echo "Error: ${e}"
+    // Set currentBuild.result as FAILURE if there is an error at any point
+    currentBuild.result = 'FAILURE'
```
mythi: you're removing this later so what's the purpose of this commit?

Jenkinsfile
```diff
@@ -189,6 +189,17 @@ try {
...
+} catch (Exception e) {
+    echo "Error: ${e}"
+    if (is_pr) {
+        // GH API cant take more than 140 chars in status msg so lets truncate.
+        def _msg = e.getMessage().take(120)
+        setGitHubPullRequestStatus state: 'FAILURE', context: "${env.JOB_NAME}", message: "Build failed: ${_msg}"
+    }
+    throw e
+} finally {
+    echo "Finally: build result is ${currentBuild.result}"
```
mythi: later you add a check that ```${currentBuild.result}``` could also be null. check its nullness before echo

Jenkinsfile
```diff
@@ -183,15 +183,20 @@ try {
...
             }
         }
     } // if testinfo_sumz
 
 } catch (Exception e) {
     echo "Error: ${e}"
-    // Set currentBuild.result as FAILURE if there is an error at any point
-    currentBuild.result = 'FAILURE'
+    if (currentBuild.result != 'UNSTABLE')
+        // Set currentBuild.result as FAILURE if there is an error in building
```
mythi: I'm having difficulties to follow the logic here. What possible values the result can have in here? If it's SUCCESS, what it means it becomes FAILURE? If it's FAILURE, why bother resetting it.

Jenkinsfile
skuusela: Yes this is indeed useless, I'll remove the commit.

Jenkinsfile
skuusela: Will do.

Jenkinsfile
skuusela: It can be 'null' so that's why it's set to FAILURE. And if it's UNSTABLE we wan't to keep it that way because that tells us that there wass an error in testing setup and not in the PR/master build.

skuusela: Yeah thanks, that makes more sense.

PR #97, opened 2017-03-30
installer image, integrity protection

pohly:
This PR is now ready for a final review. Actually building it is currently broken, most likely because the oe-selftest support in the CI system needs to be fixed first.

@ipuustin can you help review the security side of this?

@gvancuts @dbkinder can you help reviewing the new documentation (doc/security.rst and the new section in doc/howtos/image-install.rst)


okartau:
can submodule url be changed to https, so that builder can fetch it without using git:// as it does with other submodules

pohly:
test this please

pohly:
test this please

okartau:
CI tester does not find image because image file name is formed differently.
tester expects:
refkit-image-gateway-intel-corei7-64-2017-03-30_08-15-23-build-539.wic.xz
image file built:
refkit-image-gateway-intel-corei7-64-20170330155801.wic.xz

pohly:
On Thu, 2017-03-30 at 09:42 -0700, okartau wrote:
> CI tester does not find image because image file name is formed
> differently.
> tester expects:
> refkit-image-gateway-intel-corei7-64-2017-03-30_08-15-23-build-539.wic.xz
> image file built:
> refkit-image-gateway-intel-corei7-64-20170330155801.wic.xz

Does the post-build test include refkit-ci.inc in auto.conf?

There is a suspicious 10 minute delay when testing starts, for example:

00:42:32.434 2017-03-30 15:57:53,559 - selftest - INFO - Adding: "include bblayers.inc" in bblayers.conf
00:42:32.434 2017-03-30 15:57:53 - test_install_fixed_password (oeqa.selftest.image-installer.ImageInstaller)
00:52:42.780 Test image installation under qemu without virtual TPM, using a fixed password ... Loading cache...done.
00:52:42.780 rootfs file: /srv/jenkins/workspace/builder-slot-0/build/tmp-glibc/deploy/images/intel-corei7-64/refkit-installer-image-intel-corei7-64.wic

That test setup should be faster.

My theory is that actual build and post-build testing are not set up
alike enough, causing the file names to change, and therefore all images
get generated anew. The default in OE-core is to remove old images
(actually, I think only that is supported anymore), so that would
explain why the expected images are gone.

pohly:
On Thu, 2017-03-30 at 21:13 +0200, Patrick Ohly wrote:
> On Thu, 2017-03-30 at 09:42 -0700, okartau wrote:
> > CI tester does not find image because image file name is formed
> > differently.
> > tester expects:
> > refkit-image-gateway-intel-corei7-64-2017-03-30_08-15-23-build-539.wic.xz
> > image file built:
> > refkit-image-gateway-intel-corei7-64-20170330155801.wic.xz
> 
> Does the post-build test include refkit-ci.inc in auto.conf?
> 
> There is a suspicious 10 minute delay when testing starts, for example:
> 
> 00:42:32.434 2017-03-30 15:57:53,559 - selftest - INFO - Adding: "include bblayers.inc" in bblayers.conf
> 00:42:32.434 2017-03-30 15:57:53 - test_install_fixed_password (oeqa.selftest.image-installer.ImageInstaller)
> 00:52:42.780 Test image installation under qemu without virtual TPM, using a fixed password ... Loading cache...done.
> 00:52:42.780 rootfs file: /srv/jenkins/workspace/builder-slot-0/build/tmp-glibc/deploy/images/intel-corei7-64/refkit-installer-image-intel-corei7-64.wic
> 
> That test setup should be faster.
> 
> My theory is that actual build and post-build testing are not set up
> alike enough, causing the file names to change, and therefore all images
> get generated anew. The default in OE-core is to remove old images
> (actually, I think only that is supported anymore), so that would
> explain why the expected images are gone.

I think I found it:

docker/build-project.sh:export BUILD_ID=${CI_BUILD_ID}

I bet that part is missing for oe-selftest invocations, and thus
CI_BUILD_ID gets ignored, causing exactly the behavior I described
above.

pohly:
test this please

pohly:
retest this please
test this please

okartau:
test this please

okartau:
it's good to have BUILD_ID differently in test scripts than build scripts, I explained it in bug about BUILD_ID, so that part should be removed from this patch set. 
A post-build tester should not become builder.

pohly:
test this please

pohly:
test this please

pohly:
test this please

pohly:
Rebased and removed as many local workarounds as possible in the foreseeable future. The remaining ones (like enabling the swtpm spawning through qemu) will take longer to resolve upstream.

pohly:
test this please

doc/howtos/image-install.rst
```diff
@@ -86,3 +86,52 @@ includes a checksum verification, you can also use the traditional :command:`dd`
 
 Unplug the removable media from your development system and you're ready to plug
 it into your target system.
+
+Installing to internal Media
```
pohly: I did that here to be consistent with capitalization in headers elsewhere ("major" words in uppercase, adjectives lowercase - not sure whether that's exactly the rule).

I don't have a strong preference regarding media vs. medium. I used singular when talking about one particular instance and plural when there could be more than one.

doc/howtos/image-install.rst
```diff
@@ -86,3 +86,52 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+============================
+
+With the approach above, the OS always boots from the removable
+media. To install the OS to the internal media, one can build the
```
pohly: Makes sense.

doc/howtos/image-install.rst
```diff
@@ -86,3 +86,52 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+* :file:`meta-refkit/classes/image-installer.bbclass`
+
+* :file:`meta-refkit/recipes-image/images/refkit-installer-image.bb`
+
+The existing :file:`refkit-installer-image.bb` is just an example. The
+expected usage in production is that a custom installer image will
+be created which automatically installs the OS.
+
+There are also example scripts under ``doc/howtos/image-installer``
+that show how to do installation with a TPM in a qemu virtual
```
pohly: Yes.

doc/howtos/image-install.rst
```diff
@@ -86,3 +86,52 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+machine. On a build host, create a build environment for building
+without Docker as explained in the README.rst, then enter::
```
pohly: Okay. I simply wasn't sure how to link to other documentation.

doc/security.rst
```diff
@@ -0,0 +1,184 @@
...
+example for an image that is built with dm-verity enabled.
+
+In development mode (= including :file:`refkit-development.inc`), a
+key provided in :file:`meta-refkit/files/dm-verity/private.pem` is
+used for signing, which is obviously not secure.
+
+Real products must use their own key. Those can be created for
+example with::
+
+  openssl genrsa -aes128 -passout pass:<your key password> -out private.pem 4096
```
pohly: No particular reason. I just hadn't noticed that particular aspect ;-} I'll try genpkey.

doc/security.rst
```diff
@@ -0,0 +1,184 @@
...
+.. _recover from sudden power loss: https://www.youtube.com/watch?v=N8V0W0p3YBU&t=1115
+
+Whole-disk encryption protects against offline attacks as a side
+effect, assuming that the attacker cannot gain access to the secret
+key, because blocks cannot be decrypted/encrypted and thus files
+cannot be accessed and modified without the secret key.
+
+Whole-disk encryption is currently supported when installing to
+internal media using the ``image-installer`` command. See
+howtos/image-install.rst for information about this image installation
```
pohly: Okay.

doc/security.rst
```diff
@@ -0,0 +1,184 @@
...
+blocks are required for mounting the partition. This could be avoided
+by setting up ``dm-crypt`` directly, which currently is not supported
+out-of-the-box.
+
+.. _LUKS: https://gitlab.com/cryptsetup/cryptsetup/blob/master/README.md
+
+To unlock the LUKS partition, a secret key is
+needed. ``image-installer`` and the corresponding
+:file:`meta-refkit/recipes-image/images/initramfs-framework-refkit-luks.bb`
+can use a fixed passphrase. To use that, invoke image-installer with::
```
pohly: Okay.

meta-refkit/classes/image-installer.bbclass
```diff
@@ -0,0 +1,370 @@
+# This class can be inherited by an arbitrary image recipe to install
+# the "image-installer" command plus one or more image files into the
+# image.
+#
+# Basically that turns the image using image-installer.bbclass into an
+# "installer image" which can be booted from a removal media to
```
ipuustin: removal -> removable

ipuustin: Btw, commit msg says "Ostro OS OX"?

meta-refkit/classes/image-installer.bbclass
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    info () {
+        echo >/dev/tty "$*"
+    }
+}
+
+# Cleanup handling, common code for selecting among different input and output
+# options interactively, etc.
+INSTALLER_UTILS[shellcheck] = "sh"
```
ipuustin: Did you run this through shellcheck? It reports ~8 warnings and style issues, nothing major though.

meta-refkit/classes/image-installer.bbclass
```diff
@@ -0,0 +1,370 @@
...
+        info "Boot device is $root_device."
+        local exclude=
+        if [ "$root_device" ]; then
+            # lsblk knows about the device topology and orders accordingly
+            # when the NAME column is enabled. Here we are relying on the
+            # fact that dependent items appear below the disk they are
+            # rooted in.
+            local disk=
+            # shellcheck disable=SC2034
+            exclude=$( lsblk --output NAME,KNAME,TYPE --ascii --noheadings | while read name kname type; do
```
ipuustin: Is `lsblk` assumed to be provided by the host?

meta-refkit/recipes-image/images/refkit-installer-image.bb
```diff
@@ -0,0 +1,188 @@
...
+            pnum=$(expr $pnum + 1)
+            gdisk_pnum=$(expr $gdisk_pnum + 1)
+        done
+        cleanup_install_image
+    }
+}
+INSTALLER_INSTALL ?= "${REFKIT_INSTALLER_UEFI_COMBO}"
+
+INSTALLER_RDEPENDS_append = " \
+    dosfstools \
```
ipuustin: GPLv3

meta-refkit/classes/image-installer.bbclass
pohly: I did at some point, but not recently. The INSTALLER_LOGGING[shellcheck] annotation is for some non-upstream OE-core extension of the verify-bashism script, to be used with "verify-bashisms initramfs-framework-refkit-luks refkit-installer-image", which then runs shellcheck in addiditon to checking for bashisms.

meta-refkit/classes/image-installer.bbclass
pohly: That's for the target, covered by:

    INSTALLER_RDEPENDS ??= " \
        util-linux \
    "

meta-refkit/recipes-image/images/refkit-installer-image.bb
```diff
@@ -48,10 +52,20 @@ REFKIT_INSTALLER_UEFI_COMBO () {
             fi
         done
         if [ ! "$partition" ]; then
-            fatal "could not identify partition #$gdisk_pnum in $output"
+            fatal "could not identify parition #$gdisk_pnum in $output"
```
ipuustin: What's this?

meta-refkit/recipes-image/images/refkit-installer-image.bb
```diff
@@ -23,10 +23,14 @@ REFKIT_INSTALLER_UEFI_COMBO () {
         rootfs=$4
         output_mounted=
         output_mountpoint=
+        output_luks=
+        LUKS_NAME=rootfs
+        LUKS_PASSWORD=refkit
```
ipuustin: There should probably be a big warning message so that no-one will use this password accidentally.

meta-refkit/recipes-image/images/files/initramfs-framework-refkit-luks
```diff
@@ -0,0 +1,52 @@
...
+    while true; do
+        if [ $(( $C * $delay )) -gt $timeout ]; then
+            fatal "LUKS root partition '$bootparam_root' not found."
+        fi
+
+        if [ -e "$bootparam_root" ]; then
+            cryptsetup isLuks "$bootparam_root" 2>/dev/null
+            case $? in
+              0) # LUKS volumne
+                if echo "$LUKS_PASSWORD" | cryptsetup open --type luks "$bootparam_root" "$LUKS_NAME" --key-file -; then
```
ipuustin: What's the trailing `-`? Is it for reading the password?

meta-refkit/recipes-image/images/refkit-installer-image.bb
```diff
@@ -0,0 +1,188 @@
...
+        else
+            input_mounted=1
+        fi
+
+        # Clear all partition data on disk
+        if ! execute sgdisk -o "$output"; then
+            fatal "sgdisk $output has failed - damaged disk?"
+        fi
+
+        # Read partition description from rootfs.
```
mythi: Can we get rid of this emmc-partitions-data dependency? Currently, it's essentially ```DSK_IMAGE_LAYOUT``` with ```INT_STORAGE_ROOTFS_PARTUUID_VALUE``` replaced as the rootfs PARTUUID.

However, now we need to keep ```DSK_IMAGE_LAYOUT``` up-to-date with .wks.in just to get emmc-partitions-data correct.  Since the goal with our installer is just "take all but rootfs as is but make rootfs take all the space on the output device and make its ```PARTUUID==INT_STORAGE_ROOTFS_PARTUUID_VALUE```, could we do this runtime and just have ```INT_STORAGE_ROOTFS_PARTUUID_VALUE``` substituted when processing the script?

meta-refkit/recipes-image/images/files/initramfs-framework-refkit-luks
pohly: Yes, it makes cryptsetup read the password from stdin.

meta-refkit/recipes-image/images/refkit-installer-image.bb
pohly: Is INT_STORAGE_ROOTFS_PARTUUID_VALUE guaranteed to be the same for all images that the script will be asked to install? If yes, then it would be possible to embed the value directly in the script.

But probably that is an oversimplification, because technically INT_STORAGE_ROOTFS_PARTUUID_VALUE is a per-image variable and could be set differently.

mythi: True, it cannot be guaranteed and we should not make that limitation.  I had not checked this carefully enough and still had the mindset (don't know where that came to me) that the installer image had just the rootfs'es of the other images in it (and not the full images).

Anyway, the likely mismatch between ```DSK_IMAGE_FORMAT``` and what's in .wks.in is a problem. How about if do_uefiapp copies the "internal storage" .efi in ```EFI_<PARTUUID>``` and the rootfs PARTUUID is parsed from it.

With, e.g., ```sgdisk -p <wic>``` we can get the same partition data and the partitions carry the filesystem info.

meta-refkit/recipes-image/images/refkit-installer-image.bb
pohly: The password feature is always enabled, but never used unless explicitly enabled by setting FIXED_PASSWORD=yes when calling image-installer.

Are you suggesting to not put the code into the image-installer script by default and once enabled, print a warning? That seems counter-intuitive - after all, the feature would have been enabled intentionally in such a setup.

meta-refkit/recipes-image/images/refkit-installer-image.bb
pohly: @mythi I'm not exactly following what you are suggesting. Can we keep this as it is now and clean it up once the code gets reorganized (do_uefiapp -> meta-intel, etc.)?

mythi: We can clean it up later.

What I'm suggesting is:
- the image partition layout is read from the image and not from emmc_partitions_data
- the install target rootfs PARTUUID is read from the EFI directory name  (currently it's EFI_Internal_storage but could become ```EFI_<PARTUUID>```). Needs a change in do_uefiapp.

meta-refkit/recipes-image/images/refkit-installer-image.bb
ipuustin: Ok, it wasn't clear to me from the commit that the password is not used by default. It just worries me a lot whenever I see a hardcoded password in the sources. :-) But are the `FIXED_PASSWORD` and `LUKS_PASSWORD` settings overlapping a bit, if one is used only if another is enabled? I mean, could we just say `FIXED_PASSWORD="refkit" image-installer`?

doc/howtos/image-install.rst
```diff
@@ -86,3 +86,52 @@ includes a checksum verification, you can also use the traditional :command:`dd`
...
+and follow the instructions.
+
+Installing directly to internal media without the installer image is
+hardware-specific and not supported at the moment. The advantage of
+the installer image approach is that installation can use the main CPU
+and local hardware like a TPM chip to enable per-machine encryption.
+
+By default, the installer image contains different images that can be
+installed, as configured by the ``INSTALLER_SOURCE_IMAGES`` variable.
+It uses dm-verity to ensure that the installer image did not get
```
ipuustin: dm-verity -> dm-verify

doc/howtos/image-installer/init-tpm
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh -ex
```
ipuustin: Might it be a good idea to have these scripts in `intel-iot-refkit/scripts/image-installer` instead of `intel-iot-refkit/doc/howtos/image-installer`?

doc/howtos/image-install.rst
ipuustin: No, sorry, it was dm-verity. Thought this was a typo. :-)

meta-refkit/recipes-image/images/refkit-installer-image.bb
pohly: LUKS_PASSWORD is basically an internal constant. Its value has to match the string in initramfs-framework-refkit-luks.

FIXED_PASSWORD=yes is what enables the use of that fixed password.

We could make it so that there is no default password in initramfs-framework-refkit-luks unless the developer configures one in his build, and the image-installer script could be changed so that this fixed password has to passed via the FIXED_PASSWORD env variable. Would that be better? It's harder to use, but perhaps that is how we want it to be ;-}

doc/howtos/image-installer/init-tpm
pohly: I don't have a strong opinion about the location of these scripts. I can move them.

doc/security.rst
```diff
@@ -0,0 +1,184 @@
...
+Then during the build, ``REFKIT_DMVERITY_PRIVATE_KEY`` must be set to
+the full path of the resulting :file:`private.pem` and
+``REFKIT_DMVERITY_PASSWORD`` must be ``pass:<your key password>`` or
+something else that tells openssl how to obtain the key password (see
+``-passin`` in ``man openssl``).
+
+A real product also needs to ensure that the device boots with an
+initramfs which really activates dm-verity.
+:file:`meta-refkit/recipes-image/images/initramfs-framework-refkit-dm-verity.bb`
+in ``refkit-initramfs`` does that when ``dmverity`` is part of the
```
ipuustin: This is a bit confusing part. The initramfs module doesn't ensure that we are booting from an initramfs which activates dm-verity. It checks the dmverity signature instead.

pohly: The documentation refers to this line:

          if [ -e "$bootparam_root" ] && [ -e "$bootparam_dmverity" ]; then

It's the dmverity boot parameter which must be present to trigger the code, which must be set when building the image. Would it be better to say:

> A real product also needs to ensure that its image really activates dm-verity. This is done by
> booting with the ``dmverity`` boot parameter, which tells :file:`meta-refkit/recipes-image/images/initramfs-framework-refkit-dm-verity.bb`
> in ``refkit-initramfs`` that dm-verity is required. :file:`refkit-image.bbclass` automatically adds that boot parameter when
> building an image with dm-verity enabled. Secure Boot is required to prevent attackers from replacing the initramfs 
> and/or changing boot parameters.

doc/security.rst
```diff
@@ -0,0 +1,184 @@
...
+new OS installation has full control over the TPM and does not need to
+share it with other OS installations.
+
+When booting, the initramfs will retrieve the key and then prevent
+further access to it, using the TPM ``READ_STCLEAR`` protection
+feature for this particular slot. The effect is that even if an
+attacker gains access to the TPM at runtime, reading the key again
+gets denied by the TPM chip itself. However, the secret key is of
+course still in use by the kernel and thus might be leaked from there
+and/or files modified at runtime by an attacker will be written to
```
ipuustin: This sentence is difficult for me to parse.

doc/security.rst
```diff
@@ -0,0 +1,184 @@
...
+
+* Runtime Machine Configuration (RMC) database with additional
+  boot parameters which get activated depending on the current machine
+
+This UEFI combo application gets loaded directly by the firmware,
+without any intermediate boot loader involved. This approach is fast,
+simple and can be secured by signing the UEFI combo application.
+
+That signing currently has to be done manually as described in the
+`Ostro(TM) OS Secure Boot`_ documentation.
```
ipuustin: We should maybe copy that documentation?

ipuustin:
The meta-openembedded update in this PR is intentional?

meta-refkit/classes/image-installer.bbclass
```diff
@@ -0,0 +1,370 @@
...
+# several fragments provided by image-installer.bbclass, by the distro
+# or by the image recipe. The expected usage is that distros using
+# image-installer.bbclass will heavily customize the final script.
+#
+# Because the shell code is embedded in the recipe, regular bitbake
+# variables can be injected easily and the shell syntax is checked
+# already by the bitbake parser. The downside is that we are limited
+# to shell code that the bitbake parser understands (no bash
+# extensions, for example) and error reports about the shell syntax
+# are a bit hard to read.
```
ipuustin: I think the question is whether embedding the code inside recipe files makes it less maintainable in the long run. For example, how do you write tests for it?

meta-refkit/classes/image-installer.bbclass
```diff
@@ -0,0 +1,370 @@
...
+# - determine which disks are suitable as target, automatically
+#   filtering out the boot disk and everything that is read-only
+#   or completely unusable (removable media)
+# - determine which target to install to
+# - execute installation
+#
+# See below for the commands used by each step.
+#
+# The default installer code is free of bashisms and passed
+# checks with checkbashisms.pl and checkshell. The "local"
```
ipuustin: checkshell -> shellcheck ?

meta-refkit/classes/image-installer.bbclass
```diff
@@ -0,0 +1,370 @@
...
+# keyword (although supported by bash and dash) is avoided
+# in favor of wrapping function bodies with local variables
```
ipuustin: This says `local` is avoided, but I still see several variables with `local` keyword...

doc/security.rst
pohly: Perhaps. @mythi, are you working on enhancing support for Secure Boot in refkit?

pohly:
On Wed, 2017-04-05 at 00:38 -0700, Ismo Puustinen wrote:
> The meta-openembedded update in this PR is intentional?

Yes. It contains some of my fixes that are needed for this PR.

meta-refkit/classes/image-installer.bbclass
```diff
@@ -0,0 +1,370 @@
...
+# options interactively, etc.
+INSTALLER_UTILS[shellcheck] = "sh"
+INSTALLER_UTILS () {
+    # Skip user interaction. Can be set also in the environment before
+    # calling the script.
+    FORCE=${FORCE:-no}
+
+    # Checks whether a variable is "yes" or "no", then returns
+    # with 0 when "yes".
+    istrue () {
```
ipuustin: I guess `if [ "x$1" = "xyes" ];` won't cut it?

meta-refkit/classes/image-installer.bbclass
pohly: How would one write tests if it was a separate file? I don't think that would be different. In both cases one needs to boot into the installer image and then exercise the script in various ways. This can be done via oe-selftest.

Whether embedding vs. separate file is easier from a maintenance perspective is debatable. I found it quite useful.

meta-refkit/classes/image-installer.bbclass
pohly: That's what I had initially, but then typos like x=yess aren't caught.

meta-refkit/classes/image-installer.bbclass
pohly: I think I added that back later... probably should remove "local" again.

doc/security.rst
mythi: Yes I am. I saw this section and will be updating that part with what I have.

meta-refkit/classes/image-installer.bbclass
ipuustin: I didn't mean functional testing ("Does the installer work?"), but rather unit- or module-level tests ("Does pick_option() fail gracefully with these inputs?"). There are frameworks for testing shell scripts, like https://github.com/kward/shunit2. Since we don't have any tests for this it doesn't matter much at the moment, but the people customizing their own installer scripts might want to do it. This is not a showstopper though...

meta-refkit/classes/refkit-hash-dm-verity-key.bbclass
```diff
@@ -0,0 +1,28 @@
...
+    import hashlib
+
+    privkey = d.getVar('REFKIT_DMVERITY_PRIVATE_KEY')
+    if privkey is None:
+        bb.fatal('REFKIT_DMVERITY_PRIVATE_KEY is not set.')
+    if not os.path.isfile(privkey):
+        bb.fatal('REFKIT_DMVERITY_PRIVATE_KEY=%s is not a file.' % privkey)
+    with open(privkey, 'rb') as f:
+        data = f.read()
+    hash = hashlib.sha256(data).hexdigest()
```
ipuustin: Will fail or calculate wrong hash if file can't be opened? Indent to be part of the `with open...` block and/or check `data`.

pohly: It'll bail out with an exception if open() or read() fail, so data and hash are always valid here when execution continues. Indention doesn't matter. I prefer it this way because the hash operation doesn't depend on the file still being open.

meta-refkit/recipes-image/images/initramfs-framework-refkit-dm-verity.bb
```diff
@@ -0,0 +1,112 @@
...
+refkit_dmverity[shellcheck] = "sh"
+refkit_dmverity () {
+
+    dmverity_enabled() {
+        [ "$bootparam_dmverity" ]
+    }
+
+    dmverity_run () {
+        C=0
+        delay=${bootparam_rootdelay:-1}
```
ipuustin: Should you check for `bootparam_rootwait` too?

meta-refkit/recipes-image/images/refkit-installer-image.bb
```diff
@@ -0,0 +1,270 @@
...
+
+            if [ "$gdisk_pnum" -eq "$PART_COUNT" ]; then
+                # Make the last partition take the rest of the space
+                if ! execute sgdisk -n "$gdisk_pnum:+0:-1s" -c "$gdisk_pnum:$lname" \
+                       -t "$gdisk_pnum:$type_id" -u "${gdisk_pnum}:${uuid}" -- "$output"; then
+                    fatal "creating rootfs partition failed"
+                fi
+            else
+                if ! execute sgdisk -n "$gdisk_pnum:+0:+${size}M" -c "$gdisk_pnum:$lname" \
+                       -t "$gdisk_pnum:$type_id" -u "${gdisk_pnum}:${uuid}" -- "$output"; then
```
ipuustin: I'm bit confused about using both `$gdisk_pnum` and `${gdisk_pnum}` on this line.

pohly: That was copied verbatim from https://github.com/otcshare/ostro-os-xt/blob/master/meta-ostro-xt/recipes-xt/initrdscripts/initramfs-framework-installer/installer#L133

I think it is safe (and better) to replace ${gdisk_pnum} with $gdisk_pnum and ${uuid} with $uuid.

meta-refkit/recipes-image/images/initramfs-framework-refkit-dm-verity.bb
pohly: initramfs-framework doesn't use that parameter elsewhere either. It might make sense to allow waiting indefinitely (because the fallback is to give up), but that should be introduced across the board, not just here.

doc/security.rst
pohly: Is this better:

> The key remains in kernel memory and it is the kernel's responsibility to
> not leak it during a runtime attack. In addition, the kernel must prevent
> an attacker from modifying files at runtime, because such changes cannot
> be detected during the next boot.

meta-refkit/recipes-image/images/refkit-installer-image.bb
ipuustin: I agree. My first thought was that it was some BB substitution.

meta-refkit/classes/refkit-hash-dm-verity-key.bbclass
ipuustin: Ok. The code comment just talks handling the other error cases, but not these.

doc/security.rst
ipuustin: That's much better.

doc/security.rst
ipuustin: I think this is clearer.

meta-refkit/recipes-image/images/refkit-installer-image.bb
pohly: I've made it harder to accidentally use the fixed password, see the new "installer image: fixed password support disabled by default" commit.

pohly:
Another change - "use_encryption" was only documented in the commit message. It's now also in security.rst:

    diff --git a/doc/security.rst b/doc/security.rst
    index 32f2510..8d84772 100644
    --- a/doc/security.rst
    +++ b/doc/security.rst
    @@ -116,6 +116,7 @@ Currently, `LUKS`_ is used for the rootfs partition. Using LUKS has
     several advantages:
     
     * ``refkit-initramfs`` can detect whether a partition is encrypted
    +  and uses the partition accordingly
     
     * LUKS tools have sane default parameters for encryption
     
    @@ -132,6 +133,15 @@ out-of-the-box.
     
     .. _LUKS: https://gitlab.com/cryptsetup/cryptsetup/blob/master/README.md
     
    +By default, ``refkit-initramfs`` does not enforce the use of
    +encryption. That is because the images are meant to be usable both
    +after flashing to removable media (no encryption) and after
    +installation (optionally with encryption). If encryption is meant to
    +provide integrity protection, booting without encryption should be
    +disabled by adding the ``use_encryption`` boot parameter to the
    +``APPEND`` variable of an real product image. Otherwise an attacker
    +could replace the entire rootfs and then boot into that.
    +
     To unlock the LUKS partition, a secret key is
     needed. ``image-installer`` and the corresponding
     :file:`meta-refkit/recipes-image/images/initramfs-framework-refkit-luks.bb`
    diff --git a/openembedded-core b/openembedded-core
    --- a/openembedded-core
    +++ b/openembedded-core
    @@ -1 +1 @@
    -Subproject commit d68a86d87aa017dd0fecb0f626d22711efefcd3f
    +Subproject commit d68a86d87aa017dd0fecb0f626d22711efefcd3f-dirty

gvancuts:
Thanks @pohly 👍

pohly:
test this please

okartau:
test this please

pohly:
Last test run failed due to an unrelated problem:

    AssertionError: 1 != 0 : Errors in boot log:
     Mar 30 10:29:24 intel-corei7-64 kernel: [drm:intel_dp_aux_ch] *ERROR* dp_aux_ch not done status 0xa1450064,

rojkov:
would be nice if the PR included meta-openembedded update up to b0699e4. Then I wouldn't need to update the layer in my upcoming PR which drops python2 from all images.

pohly:
@mythi said on IRC that he wants to wait for a refkit snapshot to be tagged before merging, so I'm continuing to investigate the image rebuilding during post-selftest here.

@rojkov I've update meta-openembedded to latest master.

pohly:
test this please

pohly:
test this please

pohly:
Third times the charm... I hope.

test this please

pohly:
test this please

okartau:
test this please

pohly:
test this please

pohly:
test this please

pohly:
Finally managed to get a bitbake-diffsig between do_rootfs from the main build and post testing (note for future reference: when building from sstate, there is no sigdata file anywhere for that build; one has to use "bitbake -S none").

It is the buildhistory which causes the image rebuilding. From https://ostroproject.org/jenkins/job/intel-iot-refkit_pull-requests/631/consoleFull:

00:33:54.247 + bitbake-diffsigs tmp-glibc/stamps/intel_corei7_64-refkit-linux/refkit-initramfs/1.0-r0.do_rootfs.sigdata.ae662fb357580faa04a25a310e02e77a tmp-glibc/stamps/intel_corei7_64-refkit-linux/refkit-initramfs/1.0-r0.do_rootfs.sigdata.bc77000cdcfa37762afe751ef8cd55b9
00:33:54.247 Task dependencies changed from:
...
00:33:54.247 changed items: {'buildhistory_get_imageinfo'}
...
00:33:54.247 Dependency on variable buildhistory_get_build_id was added
00:33:54.247 Dependency on variable buildhistory_list_installed was added
00:33:54.247 Dependency on variable buildhistory_get_imagevars was added
...
00:33:54.247 Variable IMAGE_POSTPROCESS_COMMAND value changed from 'write_deploy_manifest; ' to 'write_deploy_manifest;   buildhistory_get_imageinfo ; '
00:33:54.247 Variable ROOTFS_POSTUNINSTALL_COMMAND value changed from 'write_image_manifest ;  stateless_mangle_rootfs;' to 'write_image_manifest ;    buildhistory_list_installed_image ;                                buildhistory_get_image_installed ; stateless_mangle_rootfs;'

pohly:
test this please

Removed the debug commit as the root cause is now understood, updated meta-security to current master.

Can be merged as-is if testing with newer meta-security succeeds.

pohly:
test this please

pohly:
test this please

pohly:
Build and oe-selftest worked, but aft testing didn't ("unstable" result?)

test this please

mythi:
On Wed, Apr 12, 2017 at 11:49 PM, Patrick Ohly <notifications@github.com>
wrote:

> Build and oe-selftest worked, but aft testing didn't ("unstable" result?)
>
>
> unstable means a glitch in DAFT itself (or something with the tester
boards)

skuusela:
Should be fixed now, test this please

pohly:
Several new commits in this revision of the PR:

* swtpm: bump SRCREV (temporary hack) a232869  - can be removed after next meta-security update
* qemu: replace CUSE with socket backend e270c71 - permanent, patches still under review and need updating, so not suitable for OE-core yet
* refkit-image.bbclass: use virtio under runqemu   47a6af3
* CI Jenkinsfile: pass host /dev/kvm into docker session  24a7c17
* refkit-ci.inc: conditionally enable the use of KVM   38bc505

This is working. The latest test build uses KVM and virtio.

@mythi: ready for merging, IMHO.

PR #98, opened 2017-03-30
RealSense SR300 color formats patch.

ipuustin:
Backported support for SR300 camera.


mythi:
No signs of the patch getting accepted in upstream?

ipuustin:
I don't see much review taking place, but of course they might get in eventually. The patches make the depth channel work -- with `cpp-headless` example command, you'll get non-black real depth and IR images.

What's noteworthy is that with the original non-upstreamed kernel patch (https://github.com/IntelRealSense/librealsense/blob/master/scripts/realsense-camera-formats_fc25.patch) I get still better results. With the non-upstreamed patch also `cpp-tutorial-1-depth` works, even if the required color format (z16) should already be present in the upstreamed patch. I've reported this to the patch authors.

ipuustin:
Tracked here: https://github.com/01org/linux-intel-4.9/issues/2

PR #99, opened 2017-03-31
Remove readline support from bluez5

ipuustin:
The reason behind this is to eventually remove libreadline from our images with production configuration.

### DIFF
--- a/meta-refkit/conf/distro/refkit.conf
+++ b/meta-refkit/conf/distro/refkit.conf
@@ -69,6 +69,9 @@ PACKAGECONFIG_pn-python3-pygobject ?= ""
 # pango, a graphical text rendering library, is not needed by us.
 PACKAGECONFIG_remove_pn-gstreamer1.0-plugins-base = "pango"
 
+# remove readline support from Bluez to reduce GPLv3 dependencies
+BAD_RECOMMENDATIONS += "bluez5-client"
+
 PACKAGECONFIG_remove_pn-pulseaudio = "avahi"
 
 # leave out readline, gdbm, and db
--- a/meta-refkit/recipes-connectivity/bluez5/bluez5_5%.bbappend
+++ b/meta-refkit/recipes-connectivity/bluez5/bluez5_5%.bbappend
@@ -1,2 +1,7 @@
 # libusb is not required. This removes dependency on libusb-compat.
 DEPENDS_remove = "libusb"
+
+# split bluetoothctl into a subpackage
+PACKAGES =+ "${PN}-client"
+FILES_${PN}-client = "${@bb.utils.contains('PACKAGECONFIG', 'readline', '${bindir}/bluetoothctl', '', d)}"
+RRECOMMENDS_${PN} += "${PN}-client"


ipuustin:
Hmm, seems that some qa test is expecting `gatttool` to be present. :-( Need to think about this a bit...

okartau:
test this please

ipuustin:
retest this please

ipuustin:
@juimonen Does this affect the bluez demos you have constructed?

juimonen:
I don't use the bluez test tools anymore, so it doesn't matter to me, at least.

ipuustin:
Ok. We will be missing the following binaries after this change:

    attrib/gatttool
    tools/obex-client-tool
    tools/obex-server-tool
    tools/bluetooth-player
    tools/obexctl
    tools/btmgmt

mythi:
@ipuustin are those tools/* binaries actually installed? smells like the recipe is wrong if it's not possible to install bluetoothd without the readline tools.

ipuustin:
@mythi No, bluetoothd will still be installed. We will just drop a bunch of commandline tools. Bluez recipe leaves the interactive binaries out if the `readline` PACKAGECONFIG option is not present. This of course affects "hackability" of the device, because the main way of interacting with bluetoothd will then be over D-Bus.

pohly:
Sorry if it was mentioned somewhere, but what about allowing these tools to link against libreadline and then just packaging them separately, without pulling that extra package into production images? Wasn't the recipe already doing that?

mythi:
My question was basically the same @pohly is asking.

ipuustin:
Looking at the bluez recipe now, it seems that the readline part has been fixed in two parts. If `readline` is not set, the binaries are not built and there is no readline dependency. If `readline` is set, the readline-dependent not-normally-installed binaries are put into `bluez5-noinst-tools` package. There are situations where something else might end up to the package, but our configuration has only the readline-dependent binaries there.

However, there's an oddity in the packaging. `/usr/bin/bluetoothctl` binary is present in the main `bluez5` package and links against readline. However, I'm not sure if this is a bug or a feature, because from bluez README (https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/README#n84):

	--disable-client

		Disable support for the command line client

		By default the command line client is enabled and uses the
		readline library. For specific systems where BlueZ is
		configured by other means, the command line client can be
		disabled and the dependency on readline is removed.

		The client should be placed in the main package along
		with the daemons. It is universally useful.

We could, of course, disregard the advice. We would then create `bluez5-client` package and install `bluetoothctl` there (maybe with RRECOMMENDS from `bluez5` to `bluez5-client`).

mythi:
@ipuustin my preference is bluetoothctl is moved to some other package so that the daemon can be installed without readline dependencies. With that, we can keep the tooling available and they can be installed on-demand.

ipuustin:
@mythi Ok, moved bluetoothctl into bluez5-client. `BAD_RECOMMENDS` is now set to remove the link from bluez5 to bluez5-client globally, but we could limit that to certain images if wanted. Anybody can still just set `RDEPENDS` and get the package.

mythi:
@ipuustin the setup looks good but I'd prefer this in two commits (e.g., if the bbappend part gets in oe-core we could just revert that commit).

okartau:
recent failure was CI system error... test this please

PR #100, opened 2017-04-03
CI: publish after build, use CI_BUILD_ID in tests, clean after pre-test

okartau:


pohly:
I agree that it is better to publish directly after building. Post-build tests shouldn't modify the build result, but they could do that accidentally, so it's better to stay on the safe side.

I've cherry-picked  771fab1 into PR#97, so we'll see how it works in practice.

pohly:
How hard would it be to run post testing with oe-selftest and the tests on real hardware in parallel? Just wondering, we don't know yet how slow the oe-selftesting will be and thus how much of a problem the serialization will be.

okartau:
test this please

okartau:
I have had same idea, to try after-build steps in parallel. It's not trivial. Post-build test step runs as part of in-docker block, and HW-testing runs as separate parallel block later, so it cant be re-arranged by simply putting Jenkinsfile lines in another order. But let's see what can be done.

pohly:
Testing as part of PR #97 was successful, let's merge this.

okartau:
test this please

PR #103, opened 2017-04-04
mraa: Update to version 1.6.0

skuusela:


mythi:
@skuusela I was working on the same in #104. That takes care of upm too and gets rid of .inc files so I prefer that one

skuusela:
@mythi Yeah we can close this one.

mythi:
@skuusela looks like we need to solve that hello mraa problem

PR #104, opened 2017-04-04
sensors: mraa and upm updates

mythi:
Obsoletes: #103


mythi:
@rojkov let me know if you agree with the mraa/upm recipe changes (wrt .inc usage)

rojkov:
sure, looks good to me

PR #105, opened 2017-04-05
[TEST] CI: Use USB stick emulation instead of flashing image

skuusela:
Instead of flashing the image to internal memory or sd card, use
BeagleBones USB mass storage emulation to boot the image and test it.
This will be much faster as no flashing needs to be done. There is a problem
with booting if both the internal memory and the USB emulation has same
rootfs disk uuid, so change it for testing purposes. If this will be used test device
internal memory or sd card can be just written over so uuid doesn't have to be
changed.

### DIFF
--- a/docker/tester-exec.sh
+++ b/docker/tester-exec.sh
@@ -73,7 +73,7 @@ testimg() {
   # Execute with +e to make sure that possibly created log files get
   # renamed, archived, published even when AFT or some of renaming fails
   set +e
-  daft ${DEVICE} ${FILENAME} --record
+  daft ${DEVICE} ${FILENAME} --record --emulateusb
   AFT_EXIT_CODE=$?
 
   # modify names inside TEST-*.xml files to contain device and img_name
--- a/meta-refkit/classes/refkit-image.bbclass
+++ b/meta-refkit/classes/refkit-image.bbclass
@@ -230,7 +230,7 @@ NOHDD = "1"
 # Image creation: add here the desired value for the PARTUUID of
 # the rootfs. WARNING: any change to this value will trigger a
 # rebuild (and re-sign, if enabled) of the combo EFI application.
-REMOVABLE_MEDIA_ROOTFS_PARTUUID_VALUE = "12345678-9abc-def0-0fed-cba987654321"
+REMOVABLE_MEDIA_ROOTFS_PARTUUID_VALUE = "12345678-9abc-def0-0fed-cba987654329"
 # The second value is needed for the system installed onto
 # the device's internal storage in order to mount correct rootfs
 # when an installation media is still inserted into the device.


skuusela:
Had one small error setting up DAFT for this. test this please

skuusela:
Test if this is stable. test this please

skuusela:
Test stablitiy more. test this please

docker/tester-exec.sh
```diff
@@ -73,7 +73,7 @@ testimg() {
   # Execute with +e to make sure that possibly created log files get
   # renamed, archived, published even when AFT or some of renaming fails
   set +e
-  daft ${DEVICE} ${FILENAME} --record
+  daft ${DEVICE} ${FILENAME} --record --emulateusb
```
mythi: how's the emulation implemented? Just trying understand is it 'per spec'.

meta-refkit/classes/refkit-image.bbclass
```diff
@@ -230,7 +230,7 @@ NOHDD = "1"
 # Image creation: add here the desired value for the PARTUUID of
 # the rootfs. WARNING: any change to this value will trigger a
 # rebuild (and re-sign, if enabled) of the combo EFI application.
-REMOVABLE_MEDIA_ROOTFS_PARTUUID_VALUE = "12345678-9abc-def0-0fed-cba987654321"
+REMOVABLE_MEDIA_ROOTFS_PARTUUID_VALUE = "12345678-9abc-def0-0fed-cba987654329"
```
mythi: why is this necessary? It's not explained in the commit message

docker/tester-exec.sh
skuusela: It's done using libcomposite module and Beaglebone USB OTG port. We have already been using this functionality with DUT "support image". Currently we boot the support image with DUT and then using it flash the image to be tested to internal memory or sd card. The script that does it is here: https://github.com/skuusela/DAFT/blob/ac26176caa803e56941f95f1db615f8675153282/testing_harness_image_extras/initialize_testing_harness/start_libcomposite

meta-refkit/classes/refkit-image.bbclass
skuusela: For some reason if the USB emulation image and the internal one have the same disk uuid there will be some errors when booting and the image won't work correctly.

skuusela: And if we want to use this way of testing image I could just remove sd cards from Minnowboard and nuke Joules internal emmc so there will be no conflicts. Or we can just change the uuid.

docker/tester-exec.sh
mythi: what's the time benefit with, e.g. the common image using bmaptool?

mythi: The improvement is promising but I'd prefer to keep testing on real HW (especially for Joule since it uses the internal eMMC) because that's what users would have too.

mythi: Meaning: let's not take this PR for now. We have the POC (this PR) tested so we can decide to take it later if faster throughput is needed.

skuusela: @mythi Yeah makes sense.

PR #106, opened 2017-04-06
meta-iotqa: After Bluetooth test disable Bluetooth

skuusela:
At the start of Bluetooth testing, Bluetooth is getting enabled so after
tests are done it should be disabled again so it doesn't interfere with
other tests. For some reason WiFi test takes about 2 minutes longer with
Minnowboard if Bluetooth isn't disabled after Bluetooth test is done.
Also remove a command that is only for Edison (which isn't supported).


skuusela:
Build server connection issue? test this please

okartau:
also this raises concern, can some major slowdown be triggered in production use by same events chain?

skuusela:
Indeed seems to reduce ~2 minutes of overall testing time.

PR #109, opened 2017-04-11
Drop dependency on python2

rojkov:


rojkov:
Ok, the test build failed at running the tests that check for python presence which is actually what we want. I'll update the PR not to use bbappends for tweaking PACKAGECONFIG.

rojkov:
For some reason bitbake creates empty package `libxml2-python` which the QA check reports about if `PACKAGECONFIG` is tweaked in a distro config. If bbappend is used this doesn't happen.

rojkov:
Added `INSANE_SKIP_libxml2-python =+ "build-deps"` to refkit.conf for now. Also python3 needs some `update-alternatives` magic to pass the failing tests.

meta-refkit/conf/distro/refkit.conf
```diff
@@ -74,6 +74,13 @@ PACKAGECONFIG_remove_pn-pulseaudio = "avahi"
 # leave out readline, gdbm, and db
 PACKAGECONFIG_pn-python3 = ""
 
+# drop python2 support
+PACKAGECONFIG_remove_pn-opkg-utils = "python"
+PACKAGECONFIG_remove_pn-libxml2 = "python"
+# there's a bug in bitbake that causes the build-deps QA check to check an empty package
```
ipuustin: Have you filed a bug about this?

rojkov: Yocto is such a nice thing I'm still struggling with how to formulate the bug properly. So far I've come up with this:

----
Given `localdata` is a local copy of `d` after calling `localdata.setVar('OVERRIDES', 'libxml2-python')` its method `getVar()` starts to return a whitespace (" ") for `localdata.getVar('RDEPENDS')` if `libxml2-python` is made an empty package with `PACKAGECONFIG_remove = "python"` in a bbappend

But if `libxml2-python` is made an empty package with `PACKAGECONFIG_remove_pn-libxml2 = "python"` in a distro config then `localdata.getVar('RDEPENDS')` returns " python-core"
 and that breaks the 'build-deps' QA check because it thinks that `libxml2-python` actually exists even if it's empty.

----

Can you get it?

rojkov:
Submitted https://bugzilla.yoctoproject.org/show_bug.cgi?id=11374
Hopefully it's clear enough.

rojkov:
rebased and force pushed.

mythi:
@rojkov I have oe-core and bitbake updated in master

rojkov:
Rebased and force pushed.

@mythi It seems the oe-core devs oppose the idea that python3 become python at some point in spite of what the python devs themselves believe in [1]. Should we stay with the current workaround or start adding locally maintained patches to ROS scripts to replace shebang lines?

[1] https://patchwork.openembedded.org/patch/139143/

meta-refkit/recipes-devtools/nodejs/nodejs_4.5.0.bbappend
```diff
@@ -0,0 +1,17 @@
+# NodeJS needs Python for node-gyp which is a NPM module bundled with
+# GYP together. GYP is a build system generator written in Python that
+# stands for "Generate Your Projects" and is used for building Node
+# modules written in C. Since Refkit doesn't support on-target
```
mythi: This isn't exactly true so let's be careful with the wording. Furthermore, even if refkit did not support this today, a refkit user might.

AFAICS this cripples node-gyp by keeping it installed (via node-npm) but simply drops dependencies to python. How about doing packaging such that npm and node-gyp are in their own packages.

meta-refkit/conf/distro/refkit.conf
```diff
@@ -77,6 +77,15 @@ PACKAGECONFIG_remove_pn-pulseaudio = "avahi"
...
 PACKAGECONFIG_pn-python3 = ""
 
+# drop python2 support
+PACKAGECONFIG_remove_pn-opkg-utils = "python"
+PACKAGECONFIG_remove_pn-libxml2 = "python"
+# There's a bug in bitbake that causes the build-deps QA check
+# to check an empty package:
+# https://bugzilla.yoctoproject.org/show_bug.cgi?id=11374
+INSANE_SKIP_libxml2-python =+ "build-deps"
+PACKAGECONFIG_remove_pn-libxml2-native = "python"
```
mythi: Is this -native part related to that bug?

meta-refkit/conf/distro/refkit.conf
```diff
@@ -77,6 +77,15 @@ PACKAGECONFIG_remove_pn-pulseaudio = "avahi"
 # leave out readline, gdbm, and db
 PACKAGECONFIG_pn-python3 = ""
 
+# drop python2 support
+PACKAGECONFIG_remove_pn-opkg-utils = "python"
```
mythi: Does opkg-utils get installed?

mythi:
@rojkov What's the current workaround and why local ROS patches would be needed? My thinking is #120 would be the way to start.

rojkov:
> @rojkov <https://github.com/rojkov> What's the current workaround and why
> local ROS patches would be needed? My thinking is #120
> <https://github.com/intel/intel-iot-refkit/pull/120> would be the way to
> start.
>
>
As a workaround I've added a bbappend to python3 that makes it inherit
update-alternatives and create a sym link (/usr/bin/python ->
/usr/bin/python3). Without the link all python scripts that have
/usr/bin/python (/usr/bin/env python) in their shebang line need to be
patched to reference python3 explicitly.

mythi:
On Thu, Apr 20, 2017 at 9:57 AM, Dmitry Rozhkov <notifications@github.com>
wrote:

> > @rojkov <https://github.com/rojkov> What's the current workaround and
> why
> > local ROS patches would be needed? My thinking is #120
> > <https://github.com/intel/intel-iot-refkit/pull/120> would be the way to
> > start.
> >
> >
> As a workaround I've added a bbappend to python3 that makes it inherit
> update-alternatives and create a sym link (/usr/bin/python ->
> /usr/bin/python3). Without the link all python scripts that have
> /usr/bin/python (/usr/bin/env python) in their shebang line need to be
> patched to reference python3 explicitly.
>
>
>
> OK, I was on the same page. With #120 neither of these would be needed.

Did ROS work with the update-alternatives approach btw?

meta-refkit/conf/distro/refkit.conf
rojkov: There's no such package in the image manifests. But the supported recipes check fails if it's missing from refkit-supported-recipes.txt
@pohly could you please help with this? Perhaps it gets pulled because of update-alternatives in coreutils.

rojkov:
>
> OK, I was on the same page. With #120 neither of these would be needed.
>
> Did ROS work with the update-alternatives approach btw?
>
>
At least a master ROS node got launched successfully (written in python).
Unofficially ROS supports both versions in its core modules starting from
Indigo [1].

Unfortunately OE-core has dropped python2 bindings for Boost and it forces
switching to python3 if ROS computer vision modules get installed on an
image.

[1] http://wiki.ros.org/indigo/Migration

meta-refkit/recipes-devtools/nodejs/nodejs_4.5.0.bbappend
rojkov: Submitted https://github.com/ostroproject/meta-iot-web/pull/37 introducing nodejs-gyp, but then the supported recipes check fails even though nodejs-gyp is not installed. It needs to be amended to ignore RDEPENDS for not installed packages.

rojkov: Actually nodejs-npm is not installed on any image. So, it's just the check's greediness issue.

mythi:
@rojkov hmm the lack of python2 boost bindings in oe-core sounds like an argument to accept 0bb82bd

ipuustin:
@mythi: Boost is difficult to compile so that it works for both Pythons at the same time. See  https://bugzilla.yoctoproject.org/show_bug.cgi?id=11104

pohly:
On Thu, 2017-04-20 at 00:23 -0700, Dmitry Rozhkov wrote:
> There's no such package in the image manifests. But the supported
> recipes check fails if it's missing from refkit-supported-recipes.txt
> @pohly could you please help with this? Perhaps it gets pulled because
> of update-alternatives in coreutils.

If "python" needs to be built, then it must be listed as supported, even
if it does not get installed on the images. There's no distinction
between "installed vs. not installed" target code and that can't be
added because it's simply unknown how the resulting binaries will be
used.

rojkov:
Ok, I see now. opkg's dep resolver can have different opinion on what to install indeed.
I'll put python@core back to the list of supported recipes.

rojkov:
- rebased;
- put `python@core` back to refkit-supported-recipes.txt;
- dropped unneeded overrides;
- force pushed.

rojkov:
* rebased and force pushed.

rojkov:
* rebased and force pushed.

rojkov:
* rebased and force pushed.

Arthil:
There was a problem in CI, where one of the boards froze at test phase. Should be fixed now. 
Test this please.

mythi:
@rojkov please rebase

rojkov:
* rebased and force pushed.

rojkov:
* rebased and force pushed.

PR #114, opened 2017-04-12
meta-refkit: Add test to check qemu

ibarraJhumberto:
Add a basic test to check that qemu runs with refkit. Based on
Yocto bug #11203. The test is to be add in a selftest/refkit
path, so it could be run with:

oe-selftest -r refkit.runqemu.RunqemuTests.test_boot_machine

Further selftest tests can be added in the refkit folder. Considering
this, running all refkit selftest tests could be easily done with:

oe-selftest --run-tests-by module refkit.*

Signed-off-by: Humberto Ibarra <humberto.ibarra.lopez@intel.com>

### DIFF
--- a/meta-refkit/lib/oeqa/selftest/refkit/runqemu.py
+++ b/meta-refkit/lib/oeqa/selftest/refkit/runqemu.py
@@ -39,16 +39,6 @@ class RunqemuTests(oeSelfTest):
         self.ovmf_recipe = 'ovmf'
         self.cmd_common = "runqemu ovmf wic slirp nographic serial"
 
-        # Avoid emit the same record multiple times.
-        mainlogger = logging.getLogger("BitBake.Main")
-        mainlogger.propagate = False
-
-        self.write_config(
-"""
-# 10 means 1 second
-SYSLINUX_TIMEOUT = "10"
-""")
-
         if not RunqemuTests.image_is_ready:
             RunqemuTests.deploy_dir_image = get_bb_var('DEPLOY_DIR_IMAGE')
             bitbake(self.refkit_recipe)
--- a/meta-refkit/lib/oeqa/selftest/refkit/runqemu.py
+++ b/meta-refkit/lib/oeqa/selftest/refkit/runqemu.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python
+# ex:ts=4:sw=4:sts=4:et
+# -*- tab-width: 4; c-basic-offset: 4; indent-tabs-mode: nil -*-
+#
+# Copyright (c) 2017, Intel Corporation.
+# All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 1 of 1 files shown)


meta-refkit/lib/oeqa/selftest/refkit/runqemu.py
```diff
@@ -0,0 +1,62 @@
...
+    deploy_dir_image = ''
+
+    def setUpLocal(self):
+        self.refkit_recipe = 'refkit-image-common'
+        self.ovmf_recipe = 'ovmf'
+        self.cmd_common = "runqemu ovmf wic slirp nographic serial"
+
+        # Avoid emit the same record multiple times.
+        mainlogger = logging.getLogger("BitBake.Main")
+        mainlogger.propagate = False
```
pohly: I think I understand which problem you are trying to solve here, but is this really the right place for it? Shouldn't that better be solved for all oeSelfTest instances?

meta-refkit/lib/oeqa/selftest/refkit/runqemu.py
```diff
@@ -0,0 +1,62 @@
...
+
+        self.write_config(
+"""
+# 10 means 1 second
+SYSLINUX_TIMEOUT = "10"
+""")
```
pohly: What's this SYSLINUX_TIMEOUT setting for? Does it affect  the refkit images at all?

meta-refkit/lib/oeqa/selftest/refkit/runqemu.py
```diff
@@ -0,0 +1,62 @@
...
+            RunqemuTests.deploy_dir_image = get_bb_var('DEPLOY_DIR_IMAGE')
+            bitbake(self.refkit_recipe)
+            bitbake(self.ovmf_recipe)
+            RunqemuTests.image_is_ready = True
+
+    @testcase(2001)
+    def test_boot_machine(self):
+        """Test runqemu machine"""
+        with runqemu(self.refkit_recipe, ssh=False, launch_cmd=self.cmd_common) as qemu:
+            self.assertTrue(qemu.runner.logged, "Failed: %s" % self.cmd_common)
```
pohly: What does this assert test? Successful booting?

mythi:
We're looking to make qemu as one of the "tester devices" in #94. With that and the additional fact that many of the feature test cases plan to use runqemu via selftests I feel this is kinda redundant.

meta-refkit/lib/oeqa/selftest/refkit/runqemu.py
ibarraJhumberto: This is done in the setUp method for all runqemu tests, this one and future ones too. However, I do get your point maybe it would be better to handle this a level up where it would apply for all selftest modules.

meta-refkit/lib/oeqa/selftest/refkit/runqemu.py
ibarraJhumberto: Made some tests, this does not affect refkit, will update it.

meta-refkit/lib/oeqa/selftest/refkit/runqemu.py
ibarraJhumberto: Yes, it is meant as a basic booting test which involves the steps mentioned in Yocto bug #11203.

ibarraJhumberto:
It might look redundant at the moment, but the plan would be to add more qemu tests here. Testing booting helps with test organization and isolation of failures. If you have too many things being tested in a single test, you could have problems identifying what went wrong.

Of course, when you run a large test manually you can be sure of when qemu doesn't boot. Still, automatic tests might report back less info, so it is better to not put many tasks in a single test.

ibarraJhumberto:
Updated with new commit to address comments.

mythi:
@ibarraJhumberto What kind of additional qemu tests are in your plans? Are they something that must be run qemu only? #94 adds the capability we run most runtime tests we run with 570x/minnowboard in qemu too.

mythi:
#94 replaces this

PR #115, opened 2017-04-13
doc: add QEMU howto

rojkov:
[skip-ci]

### DIFF
--- a/doc/howtos/qemu.rst
+++ b/doc/howtos/qemu.rst
@@ -1,0 +1,35 @@
+Using QEMU
+##########
+
+A Refkit wic image can be tested with QEMU as is. For that make sure that the
+`ovmf` target has been built::
+
+  $ bitbake ovmf
+
+Then use the command `runqemu` to start a QEMU virtual machine::
+
+  $ runqemu ovmf refkit-image-common wic
+
+This command will ask for superuser permissions, will create a TAP network
+interface for the VM automatically and will expose graphical output via a
+VNC server listening on localhost.
+
+.. note:: Make sure you run a DHCP server properly configured to listen to
+   the dynamically created TAP interface if you need working connectivity.
+
+While TAP interfaces perform best an easier solution would be to use SLIRP
+interfaces if all you need is the ability to ssh to your VM. They don't
+require having superuser permissions and enable QEMU's built-in DHCP-server.
+The following command will provide working connectivity without any
+additional setup::
+
+  $ runqemu ovmf refkit-image-common wic slirp
+
+Then connect to the VM with::
+
+  $ ssh root@localhost -p 2222
+
+In order to access VM's serial console add two more options to the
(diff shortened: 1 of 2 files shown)


doc/howtos/qemu.rst
```diff
@@ -0,0 +1,35 @@
...
+##########
+
+A Refkit wic image can be tested with QEMU as is. For that make sure that the
+`ovmf` target has been built::
+
+  $ bitbake ovmf
+
+Then use the command `runqemu` to start a QEMU virtual machine::
+
+  $ runqemu runqemu ovmf refkit-image-common wic
```
mythi: copy-paste error with all ```$ runqemu``` examples it seems: runqemu is written twice.

rojkov:
> +
> +  $ runqemu runqemu ovmf refkit-image-common wic
>
> copy-paste error with all $ runqemu examples it seems: runqemu is written
> twice.
>
>
Right, too much thoughtless yanking in vim. Fixed.

pohly:
Performance under QEMU is better with KVM enabled.

This can be done with:

    BB_ENV_EXTRAWHITE="$BB_ENV_EXTRAWHITE QEMU_USE_KVM" QEMU_USE_KVM=intel-corei7-64 {oe-selftest|runqem} ...

Or it can go into local.conf.sample, similar to this entry in refkit-ci.inc that I am currently testing in PR #97:

    # If we can access /dev/kvm, then let's use it.
    QEMU_USE_KVM = "${@ 'intel-corei7-64 intel-core2-32 qemux86 qemux86-64' if os.access('/dev/kvm', os.R_OK|os.W_OK) else '' }"

What do you guys think? This PR looks like the right place to add it.

mythi:
I'd prefer the local.conf.sample option since it's easier from the user perspective.

OTOH, would that not be appropriate in the machine configs?

pohly:
Choosing KVM changes the machine type. That's probably not something that should be done by default, i.e. not in the machine config.

rojkov:
I'd second local.conf.sample. Will update the PR if there're no objections.

mythi:
@rojkov can you rebase this and make it complete? We could start with ```QEMU_USE_KVM``` disabled by default (even if it's value is conditional).

rojkov:
* rebased;
* added `QEMU_USE_KVM` to local.conf.sample (disabled by default);
* force pushed.

doc/howtos/qemu.rst
```diff
@@ -0,0 +1,35 @@
+Using QEMU
+##########
+
+A Refkit wic image can be tested with QEMU as is. For that make sure that the
+`ovmf` target has been built::
+
+  $ bitbake ovmf
```
mythi: FWIW, meta-intel added ovmf in ```EXTRA_IMAGEDEPENDS``` so it should be available. But I guess this won't hurt.

PR #116, opened 2017-04-18
Clean up and enablement of single-node Bluetooth test cases for extensive testing

jairglez:
This PR adds manifests for extensive tests per profile and cleans up and includes single-node Bluetooth test cases inside them (this is tracked on Yocto bug #11255).

The purpose for the new -ext.manifest files is to contain these and other extensive feature tests, which could be executed with a lower cadence than the sanity tests.

In order to include these tests on the CI, a bi-weekly job would need to be scheduled, and a mechanism to execute the extensive manifests' test cases would be required.

### DIFF
--- a/meta-iotqa/conf/test/refkit-image-common-extensive.manifest
+++ b/meta-iotqa/conf/test/refkit-image-common-extensive.manifest
@@ -1,1 +1,3 @@
 # Extensive tests for common profile
+oeqa.runtime.bluetooth.comm_bt_command
+oeqa.runtime.bluetooth.comm_bt_6lowpan
--- a/meta-iotqa/conf/test/refkit-image-common-extensive.manifest
+++ b/meta-iotqa/conf/test/refkit-image-common-extensive.manifest
@@ -1,0 +1,1 @@
+# Extensive tests for common profile
--- a/meta-iotqa/conf/test/refkit-image-computervision-extensive.manifest
+++ b/meta-iotqa/conf/test/refkit-image-computervision-extensive.manifest
@@ -1,0 +1,1 @@
+# Extensive tests for computervision profile
--- a/meta-iotqa/conf/test/refkit-image-gateway-extensive.manifest
+++ b/meta-iotqa/conf/test/refkit-image-gateway-extensive.manifest
@@ -1,0 +1,1 @@
+# Extensive tests for gateway profile
--- a/meta-iotqa/lib/oeqa/runtime/bluetooth/bluetooth.py
+++ b/meta-iotqa/lib/oeqa/runtime/bluetooth/bluetooth.py
@@ -1,13 +1,11 @@
 import time
 import os
-import string
 from oeqa.utils.helper import shell_cmd_timeout
 
+
 class BTFunction(object):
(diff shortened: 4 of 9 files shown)


meta-iotqa/conf/test/refkit-image-common-extensive.manifest
```diff
@@ -1 +1,3 @@
 # Extensive tests for common profile
+oeqa.runtime.bluetooth.comm_bt_command
```
mythi: are these taking so long time they're not usable part of regular testing?

jairglez: These two classes take less than a minute to complete. However, there are still more Bluetooth multi-node test cases left to enable requiring a second device for testing connection. All these times will start adding up when more test cases are enabled for this and other features and executed for each profile image if they are common between them.

Also, I didn't include the comm_bt_stability test cases, as they take about half an hour to complete considering they are stress tests. At this moment, I'm not sure if they could be useful in the future, but if they are not, it'd probably be a good idea to delete them.

jairglez: I also re-executed the tests using the latest build from master and noticed bluetoothctl (which the comm_bt_command tests use) is missing from the image, seemingly due to https://github.com/intel/intel-iot-refkit/commit/0dc352e992414fd5cff0a49964649b5e4b1d77c0 and now those tests are failing. According to this, I will need to change them to not depend on it.

ipuustin: If you need `bluetoothctl` on some testing image, add `BAD_RECOMMENDATIONS_remove = "bluez5-client"` to your config.

jairglez: Thanks @ipuustin, I tested it and it worked. The main issue with this approach, however, is that the objective is to execute these tests for all current profile images, so the configuration would need to be adjusted for all of them.

mythi: @jairglez if you install it in ```REFKIT_IMAGE_EXTRA_INSTALL``` it gets pulled in all profiles.

jairglez: @mythi, would it be ok then to enable bluetoothctl for the generated images in the CI? If that's the case, where is the best location to set the REFKIT_IMAGE_EXTRA_INSTALL variable? Also, considering that these tests take less than 1 minute, but they will start adding up with other tests, you still would like me to move them to regular sanity testing manifests? How much total time for runtime test execution per profile and platform would you consider the ideal at this time?

mythi: @jairglez actually, adding it in packagegroup-common-test.bb would probably be the best option. the other alternative is to do it like @ipuustin mentioned but packagegroup-common-test.bb makes toggling all test content on/off easier.

jairglez:
Replaced by https://github.com/intel/intel-iot-refkit/pull/143

PR #119, opened 2017-04-18
[TEST BUILD] ostree-upgradable image support

klihub:
Try to test-build in the CI infra to see the amount of damage we've managed to inflict...

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -35,3 +35,8 @@
 [submodule "meta-security"]
 	path = meta-security
 	url = https://git.yoctoproject.org/git/meta-security
+[submodule "meta-flatpak"]
+	path = meta-flatpak
+	url = https://github.com/klihub/meta-flatpak
+	branch = intel-iot-refkit/flatpak
+        ignore = untracked
--- a/.gitmodules
+++ b/.gitmodules
@@ -40,3 +40,6 @@
 	url = https://github.com/klihub/meta-flatpak
 	branch = intel-iot-refkit/flatpak
         ignore = untracked
+[submodule "meta-gplv2"]
+	path = meta-gplv2
+	url = https://git.yoctoproject.org/git/meta-gplv2
--- a/.gitmodules
+++ b/.gitmodules
@@ -38,7 +38,7 @@
 [submodule "meta-flatpak"]
 	path = meta-flatpak
 	url = https://github.com/klihub/meta-flatpak
-	branch = intel-iot-refkit/flatpak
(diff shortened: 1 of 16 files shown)


okartau:
test this please

okartau:
test this please

klihub:
The tests got stuck because the swtpm test script could not find netstat, which it fails to check and detect, in which case (if the swtpm process the script would like to connect to keeps running) the script essentially falls into an infinite loop.

The immediate reason why netstat was not found is that it was not on the PATH of the test script. My current guess is that the reason for this is related to the usrmerge patches: the script has the same (recipe-sysroot-native)/usr/bin and (recipe-sysroot-native)/usr/sbin twice on its PATH which looks suspicious, while at the same time the native recipes ignore usrmerge and install stuff in a split /-/usr fashion.

So I suspect that usrmerge somehow gets propagated to the paths which effect how the tests are run, in practice dropping /bin and /sbin from the PATH, while the actual installation happens ignoring usrmerge exactly to these dropped locations. Sigh...

Ditto for flatpak build.

klihub:
Let's see if a quick workaround gets rid of this problem...

klihub:
Well, I guess I can call it a day, then...

avalluri:
@klihub renaming the 'swtpm-wrappers.bb' recipe to 'swtpm-wrappers-native.bb'(and related changes in refkit.conf, refkit-ci.inc, image-installer.py, init-tpm, runqemu-installer, and runqeum-insternal-disk)  fixes the issue. I think bitbake's convention is that all native recipes should be suffixed with '-native', then its is treating differently.

pohly:
On Sun, 2017-04-23 at 15:01 -0700, Amarnath Valluri wrote:
> @klihub renaming the 'swtpm-wrappers.bb' recipe to
> 'swtpm-wrappers-native.bb'(and related changes in refkit.conf,
> refkit-ci.inc, image-installer.py, init-tpm, runqemu-installer, and
> runqeum-insternal-disk) fixes the issue. I think bitbake's convention
> is that all native recipes should be suffixed with '-native', then its
> is treating differently.

There's an "inherit native" in swtpm-wrappers.bb. If that's not enough,
then the class should enforce that convention or at least warn about it.

Sounds like this needs to be brought up on the OE-core mailing list. Has
that been done while I was on vacation?

avalluri:
@pohly, I realized this yesterday, so I did not bring this yet to oe-core list.

mythi:
replaced by #208

PR #122, opened 2017-04-19
[TEST BUILD] Flatpak application support rebased on latest intel/master.

klihub:


klihub:
test this please

okartau:
test this please

okartau:
test this please

klihub:
The tests got stuck because the swtpm test script could not find netstat, which it fails to check and detect, in which case (if the swtpm process the script would like to connect to keeps running) the script essentially falls into an infinite loop.

The immediate reason why netstat was not found is that it was not on the PATH of the test script. My current guess is that the reason for this is related to the usrmerge patches: the script has the same (recipe-sysroot-native)/usr/bin and (recipe-sysroot-native)/usr/sbin twice on its PATH which looks suspicious, while at the same time the native recipes ignore usrmerge and install stuff in a split /-/usr fashion.

So I suspect that usrmerge somehow gets propagated to the paths which effect how the tests are run, in practice dropping /bin and /sbin from the PATH, while the actual installation happens ignoring usrmerge exactly to these dropped locations. Sigh...

klihub:
test this please

mythi:
test this please

mythi:
replaced by #181

PR #124, opened 2017-04-20
CI Jenkinsfile: keep build workspaces with later off-build-path deletion

okartau:
"Keep workspaces" brings 2 improvements:
 - workspaces from few previous builds are present, allowing
   to check for state even after next build in same worker started;
 - deletion of older build trees does not take time of next build,
   it is performed by async-runned cleanup job.

Deletion of older workspaces is done after build in parallel with tester jobs,
scheduled in the same builder which performed build stage.

Additionally, this PR has smaller simplification in 1st commit,
2 simplifications related to single-MACHINE build in commits 4 and 5,
one status print shortening in commit 6,
one variable-defined-once improvement in commit 7.
Commits 8 and 9 deal with post-build test recently surfaced needs.


okartau:
mythi pointed out new cleanup work here overlaps with work being done in #94, so I will hold back and re-do those changes not covered by changes in #94

PR #125, opened 2017-04-20
net-tools_%.bbappend: don't break class-target build.

klihub:
Make the class target empty disable_i18n function more shell-friendly
by adding a shell-noop. Otherwise the build might abort with a shell
error along th elines of "error near unexpected token '}'".

### DIFF
--- a/meta-refkit/recipes-extended/net-tools/net-tools_%.bbappend
+++ b/meta-refkit/recipes-extended/net-tools/net-tools_%.bbappend
@@ -4,6 +4,7 @@ disable_i18n() {
        sed -i -e 's/^I18N=1/# I18N=1/' ${S}/config.make
 }
 disable_i18n_class-target () {
+  :
 }
 
 do_configure_append () {


mythi:
@klihub the bbappend in our repo is a workaround (see 990672dc7fb02112f93da4bcae07ad5e7bc1f7dc). Looks like to original patch sent to OE-core had it right. Instead accepting this change, we should actually revert 990672dc7fb02112f93da4bcae07ad5e7bc1f7dc because we now have the right fix via OE-core.

klihub:
Ok. I close this one then.

mythi:
@klihub could you send the revert PR?

klihub:
That said, it remains to be seen if that will break exactly the same way as this. I think it is some bitbake [pre|post]processing magic which is breaking the patch for me. Regardless of the bbappend having the function definition without the extra \n, once bitbake is done wiht its mangling and spitting out run.* scripts the same functions ends up with an extra \n in the body.

The only difference now will be that it does not get pulled in via a bbappend which might do the trick. However if the extra \n will get added also as part of the basic .bb processing then it will be still broken for me... and it will require a :-nop, a true or something else in the body to prevent breakage.

mythi:
The commit in the repo has it right AFAICS: http://git.openembedded.org/openembedded-core/commit/?id=508163bef169cf0d9aa97e73c02d1ecc68480e91

PR #126, opened 2017-04-21
qemu: Do not limit TPM CUSE driver removal to only native builds

avalluri:
Usage of SRC_URI_remove_pn-qemu-native, limits the patch removal only for native
builds, where it should apply even for nativesdk builds.

Signed-off-by: Amarnath Valluri <amarnath.valluri@intel.com>

### DIFF
--- a/meta-refkit/recipes-devtools/qemu/qemu_%.bbappend
+++ b/meta-refkit/recipes-devtools/qemu/qemu_%.bbappend
@@ -1,7 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/qemu:"
 
 # Replace CUSE backend with socket backend.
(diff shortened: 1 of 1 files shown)


meta-refkit/recipes-devtools/qemu/qemu_%.bbappend
```diff
@@ -1,7 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/qemu:"
 
 # Replace CUSE backend with socket backend.
-SRC_URI_remove_pn-qemu-native = " \
+SRC_URI_remove = " \
```
pohly: Yes, that's better. The pn-qemu-native was a left over from the previous experimental approach of removing it in refkit.conf.

mythi:
test this please

PR #113, opened 2017-04-12
Add nftables support.

ipuustin:
Move Refkit firewall support from iptables to nftables. The benefit of the new framework would be file-based configuration of firewall rules and the ability to assign interfaces to zones. Also, in the framework the firewall rules are executed as a single ruleset, meaning ports might be opened already before the port user is started up. The firewall transactions are now completely atomic.

PR #127, opened 2017-04-24
Add and enable iptables test

skuusela:
The test will try to drop and reject SSH connection using iptables.


mythi:
@skuusela we're looking into moving to nftables (#113) 

@ipuustin

ipuustin:
@mythi I think the key here is to figure out how to test firewall (packets being dropped etc.) without causing too long timeouts in the system. The testing should then be moved easily enough on nftables. @skuusela, add `test-build` label?

PR #57, opened 2017-03-02
Add Caffe and a precompiled imagenet model

ipuustin:
Add `caffe` and its dependencies to RefKit. The original recipes for `caffe`, `hdf5`, `lmdb`, `openblas`, and `snappy` from meta-convnet layer (https://github.com/norxander/meta-convnet, MIT and GPLv2 licenses), but have been modified quite much.

Use Caffe from the opencl git branch, meaning that it can be accelerated using OpenCL-compatible GPUs. Remove some of Caffe's Python dependencies (such as https://github.com/scikit-image/scikit-image) and replace them with easier-to-package alternatives (Pillow and ImageIO). Move all tools to use Python 3 instead of Python 2.

Package Caffe's reference imagenet model for classification of images. Add some Python demo classification code which uses the model.

PR #130, opened 2017-04-26
meta-refkit-extra: add new layer and computer vision demo content there.

ipuustin:
Add meta-refkit-extra layer for storing demos and extra content. Have the layer initially in disabled state. See https://lists.yoctoproject.org/pipermail/intel-iot-refkit/2017-March/000005.html for discussion.


meta-refkit/conf/bblayers.conf.sample
```diff
@@ -35,6 +35,10 @@ REFKIT_LAYERS += "##OEROOT##/../meta-openembedded/meta-python"
 # REFKIT_LAYERS += "##OEROOT##/../meta-openembedded/meta-xfce"
 # REFKIT_LAYERS += "##OEROOT##/../meta-openembedded/toolchain-layer"
 
+# Demo layer, disabled by default. To enable, uncomment the line below
+# and increment LCONF_VERSION by one.
+# REFKIT_LAYERS += "##OEROOT##/../meta-refkit-extra"
```
pohly: I suspect that this comment might become misleading once bblayers.conf.sample turns into conf/bblayers.conf. A developer editing that bblayers.conf the might uncomment the line and bump LCONF_VERSION, which wouldn't be right.

Do we need this change in bblayers.conf.sample at all?

A developer who wants the extra layer can do "bitbake-layers add-layer <path to meta-refkit-extra>".

mythi:
On Wed, Apr 26, 2017 at 11:02 AM, Patrick Ohly wrote:

>
> Do we need this change in bblayers.conf.sample at all?
>
> A developer who wants the extra layer can do "bitbake-layers add-layer ".
>
>
+1

meta-refkit/conf/bblayers.conf.sample
ipuustin: Ok, I'll remove the comment.

mythi:
@ipuustin will you add #57 content in this PR too?

ipuustin:
@mythi yes, added now. The only "refkit-affecting" change is the OpenCV Python 3 dependency change.

mythi:
@ipuustin thanks, this looks good to me. Let's take it once the documentation is finished.

ipuustin:
Ok, rebased. The docs are now in ok shape.

ipuustin:
@pohly : ok, I did the changes you suggested.

meta-refkit-extra/recipes-convnet/boost/boost_%.bbappend
```diff
@@ -0,0 +1 @@
+PACKAGECONFIG += "python"
```
mythi: @ipuustin Is this needed? IIRC you mentioned it's not needed.

ipuustin: That's still needed -- it enables Python support in boost. We no longer need to select Python 3 by editing the `user-config.jam` file, because Python 3 support is now enabled as default if it's found. See difference to the previous iteration of the same file. https://github.com/ipuustin/intel-iot-refkit/blob/165b46149a2c5a58788b91d2cac97c394e567f14/meta-refkit/recipes-convnet/boost/boost_%25.bbappend

However, it's a good question if we should move the `PACKAGECONFIG` variable to the layer distro configuration file instead of the `.bbappend` file. What do you think?

mythi: Ah, I see. ```python``` isn't enabled by default. I got confused by what you wrote earlier: "The bbappend should not be needed anymore."

pohly:
On Thu, 2017-04-27 at 21:43 -0700, Ismo Puustinen wrote:
> However, it's a good question if we should move the PACKAGECONFIG
> variable to the layer distro configuration file instead of
> the .bbappend file. What do you think?

Yes, please move it.

One of the "Yocto Compatibility 2.0" criteria is that a layer must not
change the build result merely because it was added to bblayers.conf,
because adding or removing layers to enable or disable something is
awkward.

I intend to add an oe-selftest which calls yocto-compat-layer on our
layers.

mythi:
On Fri, Apr 28, 2017 at 9:21 AM, Patrick Ohly wrote:


> One of the "Yocto Compatibility 2.0" criteria is that a layer must not
> change the build result merely because it was added to bblayers.conf,
> because adding or removing layers to enable or disable something is
> awkward.
>
>
Is the compatibility scope limited to adding/removing the layer? In order
to use the layer properly the user must change the config but is it so that
it does not count?

pohly:
On Thu, 2017-04-27 at 23:38 -0700, Mikko Ylinen wrote:
> On Fri, Apr 28, 2017 at 9:21 AM, Patrick Ohly wrote:
> 
> 
> > One of the "Yocto Compatibility 2.0" criteria is that a layer must
> not
> > change the build result merely because it was added to
> bblayers.conf,
> > because adding or removing layers to enable or disable something is
> > awkward.
> >
> >
> Is the compatibility scope limited to adding/removing the layer?

Merely adding/removing must not have an effect.

>  In order
> to use the layer properly the user must change the config

For "the config" = local.conf or the distro config, yes.

>  but is it so that
> it does not count?

I'm not sure anymore what "it" is here.

mythi:
On Fri, Apr 28, 2017 at 9:44 AM, Patrick Ohly <notifications@github.com>
wrote:

> On Thu, 2017-04-27 at 23:38 -0700, Mikko Ylinen wrote:
>
>

> > In order
> > to use the layer properly the user must change the config
>
> For "the config" = local.conf or the distro config, yes.
>
> > but is it so that
> > it does not count?
>
>
the mandatory local.conf change

ipuustin:
I think @mythi means that this does not matter much because in order to properly use the layer, the user needs to do two things: add the layer and include the configuration file. This will anyway cause the build result to change.

I like the concept of separating policy (configuration files) and mechanism (recipes), even though it makes `recipes-convnet` directory not to be "self-contained" anymore. I can do a PR which moves the PACKAGECONFIG to the configuration file.

ipuustin:
PR here: https://github.com/intel/intel-iot-refkit/pull/133

PR #133, opened 2017-04-28
meta-refkit-extra: move Boost package config to configuration file.

ipuustin:
Do not have Boost PACKAGE_CONFIG variable set in a .bbappend file, because adding or removing layers should not affect the build result. Have the configuration setting instead in layer's distro config file.

### DIFF
--- a/meta-refkit-extra/conf/distro/include/refkit-extra.conf
+++ b/meta-refkit-extra/conf/distro/include/refkit-extra.conf
@@ -4,3 +4,6 @@ SUPPORTED_RECIPES_append = " \
      ${META_REFKIT_EXTRA_BASE}/conf/distro/include/refkit-extra-supported-recipes.txt \
 "
 SUPPORTED_RECIPES[refkit-extra-supported-recipes.txt] = "extra"
+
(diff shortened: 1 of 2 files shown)


meta-refkit-extra/conf/distro/include/refkit-extra.conf
```diff
@@ -4,3 +4,6 @@ SUPPORTED_RECIPES_append = " \
      ${META_REFKIT_EXTRA_BASE}/conf/distro/include/refkit-extra-supported-recipes.txt \
 "
 SUPPORTED_RECIPES[refkit-extra-supported-recipes.txt] = "extra"
+
+# Caffe needs Boost Python 3 bindings. Enable them in Boost.
+PACKAGECONFIG_pn-boost += "python"
```
pohly: I think you may need PACKAGECONFIG_append_pn-boost = " python" here. 

PACKAGECONFIG_pn-boost overrides the weaker PACKAGECONFIG default instead of extending it.

ipuustin: Ok. I checked that `python` was part of the PACKAGECONFIG, but I didn't check if something was missing there. I'll fix the patch.

PR #134, opened 2017-05-02
intel-iot-refkit: submodule updates

mythi:


mythi:
test this please

ipuustin:
It seems that openembedded-core update is breaking clang. It's this line and the line below it which don't succeed: https://github.com/kraj/meta-clang/blob/master/recipes-devtools/clang/clang_git.bb#L116 . I need to dig in a bit to see what has changed so that `make` fails for native.

mythi:
test this please

PR #135, opened 2017-05-02
meta-refkit-extra testing

pohly:
We can use oe-selftest to cover some aspects of meta-refkit-extra.


meta-refkit/lib/oeqa/selftest/meta-refkit-extra.py
```diff
@@ -0,0 +1,92 @@
...
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
```
ipuustin: It might be more consistent to use MIT license, because the layer is overall MIT-licensed. I see that other scripts have been licensed under GPLv2 though.

meta-refkit/lib/oeqa/selftest/meta-refkit-extra.py
```diff
@@ -0,0 +1,92 @@
...
+class MetaRefkitExtra(oeSelfTest):
+    """
+    These tests must run in a build configuration that does *not* have
+    meta-refkit-extra.
+    """
+
+    def setUpLocal(self):
+        """
+        This code is executed before each test method.
+        It verifies that meta-refkit-extra is not active.
```
ipuustin: You could maybe just skip adding the layer if it is already present?

meta-refkit/lib/oeqa/selftest/meta-refkit-extra.py
```diff
@@ -0,0 +1,92 @@
...
+        '''
+        Adds the meta-refkit-extra layer and enables its content.
+        Will be undone after the test completes.
+        '''
+        self.write_bblayers_config('REFKIT_LAYERS += "%s"' % self.refkit_extra_path)
+        self.write_config(self.extra_conf)
+
+    def test_yocto_compat(self):
+        """
+        Checks for 'Yocto Compatible 2.0' using the yocto-compat-layer.py tool.
```
ipuustin: You could open this up in the comment a bit. What is "Yocto Compatible 2.0" and where is it defined? What does it mean if the script succeeds?

meta-refkit/lib/oeqa/selftest/meta-refkit-extra.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        runCmd("yocto-compat-layer.py '%s'" % self.refkit_extra_path)
+
+    def test_build_computervision(self):
+        """
+        Build an image with the extra packages mentioned in meta-refkit-extra/doc/computervision.rst.
+        """
+        self.enable_refkit_extra()
+        image = 'refkit-image-common'
+        packages = 'caffe-imagenet-model python3-pyrealsense opencv'
```
ipuustin: OpenCV is not coming from this layer. Caffe should bring it in as a dependency.

meta-refkit/lib/oeqa/selftest/meta-refkit-extra.py
```diff
@@ -0,0 +1,92 @@
...
+        Checks for 'Yocto Compatible 2.0' using the yocto-compat-layer.py tool.
+        """
+        runCmd("yocto-compat-layer.py '%s'" % self.refkit_extra_path)
+
+    def test_build_computervision(self):
+        """
+        Build an image with the extra packages mentioned in meta-refkit-extra/doc/computervision.rst.
+        """
+        self.enable_refkit_extra()
+        image = 'refkit-image-common'
```
ipuustin: Why not use `refkit-image-computervision` as the baseline?

pohly:
On Wed, 2017-05-03 at 00:41 -0700, Ismo Puustinen wrote:
> It might be more consistent to use MIT license, because the layer is
> overall MIT-licensed. I see that other scripts have been licensed
> under GPLv2 though.

That's exactly the reason. By making it GPLv2, it becomes possible to
copy code from other test scripts which are under GPLv2.

pohly:
On Wed, 2017-05-03 at 00:42 -0700, Ismo Puustinen wrote:
> You could maybe just skip adding the layer if it is already present?

No, yocto-compat-layer.py depends on the layer not being present
already.

pohly:
On Wed, 2017-05-03 at 00:45 -0700, Ismo Puustinen wrote:
> You could open this up in the comment a bit. What is "Yocto Compatible
> 2.0" and where is it defined?

It's the next revision of the Yocto compliance program. What exactly it
means is under discussion and hasn't been written down. Therefore I had
to leave it open here, too.

>  What does it mean if the script succeeds?

That it hasn't found problems. How much that means in practice depends
on how the script evolves.

pohly:
On Wed, 2017-05-03 at 00:46 -0700, Ismo Puustinen wrote:
> OpenCV is not coming from this layer. Caffe should bring it in as a
> dependency.

So I should remove it here?

I was following the computervision.rst, which explicitly calls out the
need to install opencv:

    You need to have packages ``python3-pyrealsense`` and ``opencv``
    installed

Should opencv be removed there, too?

ipuustin:
@pohly : that example doesn't have Caffe, so OpenCV needs to be installed manually. It doesn't hurt to have it in your `packages` variable, it's just a no-op.

pohly:
On Wed, 2017-05-03 at 00:50 -0700, Ismo Puustinen wrote:
> Why not use refkit-image-computervision as the baseline?

Because computervision.rst didn't say that the demo only works when
using refkit-image-computervision as base, and I wanted this test to be
as fast as possible, i.e. use a smaller image.

ipuustin:
@pohly : it should work with refkit-image-common too. I was sort of expecting computer vision components to be run on refkit-image-computervision, but that was maybe a false assumption.

ipuustin:
Ok .This PR looks good to me.

mythi:
@pohly this has conflicts after my secureboot selftests were added. can you resolve that? I'd like to see re-testing with this to see how much time is spent in post-selftests after this.

pohly:
On Tue, 2017-05-09 at 01:49 -0700, Mikko Ylinen wrote:
> @pohly this has conflicts after my secureboot selftests were added.
> can you resolve that? I'd like to see re-testing with this to see how
> much time is spent in post-selftests after this.

The layer tests in PR #147 are now doing something related.

Do you want to merge this PR here before the 2.3 release, or can it wait
until we know what we want to do with PR #147?

mythi:
On Thu, May 11, 2017 at 5:42 PM, Patrick Ohly wrote:

> On Tue, 2017-05-09 at 01:49 -0700, Mikko Ylinen wrote:
> > @pohly this has conflicts after my secureboot selftests were added.
> > can you resolve that? I'd like to see re-testing with this to see how
> > much time is spent in post-selftests after this.
>
> The layer tests in PR #147 are now doing something related.
>
> Do you want to merge this PR here before the 2.3 release, or can it wait
> until we know what we want to do with PR #147?
>
>
It can/should wait post release. I was mostly curious getting the build
running
to see the post-build stats but now that I think of it again, it's not
important right now.

pohly:
Rebased on top of PR #184.

The yocto-compat-layer.py test is now in refkit_poky.py. However, since also software layers are checked for signature changes now, meta-refkit-extra does not pass anymore (due to glog_%.bbappend:DEPENDS += "gflags").

Is that perhaps a change that we want in meta-refkit-core under a df-refkit-config condition? It would also avoid some recompilation when switching between building without and with the layer.

I've not finished the actual build test locally. Let's see whether it works in the CI...

ipuustin:
It's sort of ironic that one of the key reasons we put the extra packages to `meta-refkit-extra` layer (instead of just having a `recipes-extra` or similar) was the ability to alter distro and build configuration. :-) Should we instead define a new distro feature (`df-refkit-extra-config`?) in the extras layer and then make the configuration depend on that? The idea was AFAIK allow recompiling stuff when the extras layer is added.

pohly:
On Mon, 2017-06-19 at 01:03 -0700, Ismo Puustinen wrote:
> It's sort of ironic that one of the key reasons we put the extra
> packages to meta-refkit-extra layer (instead of just having a
> recipes-extra or similar) was the ability to alter distro and build
> configuration. :-)

That option still exists. The question just is whether it's done "the
old way" (= including the layer in bblayers.conf changes the build) or
"the Yocto Compatible 2.0 way" (= via distro features and/or .inc
files).

>  Should we instead define a new distro feature
> (df-refkit-extra-config?) in the extras layer and then make the
> configuration depend on that? The idea was AFAIK allow recompiling
> stuff when the extras layer is added.

I don't think we had considered "Yocto Compatible 2.0" at the time when
we made that decision. To be honest, I don't remember much discussion
around that particular point at all. IMHO it was more about "when using
an extra layer, we can guarantee that the main layer does not
accidentally depend on extra".

I think it is at least worth reconsidering the question now.

That general question aside, regarding glog:
* is that something that also would make sense in meta-refkit-core?
* is it really reliable to change the content via DEPENDS? That glog
  compiles differently looks like a bug in the recipe to me, it should
  have an explicit PACKAGECONFIG for this where using gflags is disabled
  even if it happens to be found.

ipuustin:
Using `PACKAGECONFIG` would of course be clearer. Many software components change their build depending of what optional packages they can auto-detect during configure. Before RSS this was a clear bug, nowadays it's bit murkier because we can control the available packages.

Glog is coming from meta-openembedded. I don't really know why it's configured there the way it is configured. It probably isn't used by anything else than caffe? The real question is if we want to move configuration of packages used by `meta-refkit-extra` to `meta-refkit-core`, which creates a soft dependency (the build configuration is different because meta-refkit-extra might be included).

ipuustin:
This PR could also be rebased on master.

pohly:
On Tue, 2017-06-20 at 01:26 -0700, Ismo Puustinen wrote:
> Using PACKAGECONFIG would of course be clearer. Many software
> components change their build depending of what optional packages they
> can auto-detect during configure. Before RSS this was a clear bug,
> nowadays it's bit murkier because we can control the available
> packages.

I think it is still a bug. The situation is less likely to trigger, but
unexpected, implicit compile changes can still just as before (for
example, when a dependency changes its own dependencies).

> Glog is coming from meta-openembedded. I don't really know why it's
> configured there the way it is configured. It probably isn't used by
> anything else than caffe? The real question is if we want to move
> configuration of packages used by meta-refkit-extra to
> meta-refkit-core, which creates a soft dependency (the build
> configuration is different because meta-refkit-extra might be
> included).

I don't understand the "soft dependency" part. What I had in mind was to
have a glog_%.bbappend in meta-refkit-core which for "refkit-config"
changes the compilation in glob. Then we don't need to recompile glob
and everything that depends on it when adding meta-refkit-extra.

The reason for having that change is caffe in meta-refkit-extra, but
that's not different from building the core in a way that it is usable
by additional, external layers out-of-the-box. We can do that unless
there are reasons (performance, security, etc.) which make such a change
undesirable for other cases.

ipuustin:
That's what I meant with "soft dependency". We are carrying a configuration change in `-core` which is not needed by anything. If there was no "soft dependency", `-core` wouldn't need to know about `glog` at all. A similar case would be adding a bunch of different kernel modules to all images because someone in `-extra` might need them in some configuration. It's not a real dependency, but still causing dependent changes in `-core`.

PR #136, opened 2017-05-02
compare sstate signatures between main build and post-build

pohly:
We can check that the script changes do not break the build when
testing the PR, but we will have to merge and run a master build
before we'll know what's slowing down the master builds.

### DIFF
--- a/docker/build-project.sh
+++ b/docker/build-project.sh
@@ -130,6 +130,10 @@ else
   _bitbake_targets="$BUILD_TARGET"
 fi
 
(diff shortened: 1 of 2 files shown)


pohly:
As expected, bitbake-diffsigs found no changes while testing the PR.

Agreed to merge this and then do a master build?

okartau:
ok from my side, should not be risk as affects CI only.
Is the idea these tests remain, or are added only temporarily?

pohly:
I'm undecided whether we should keep the checks. They don't take too long, but every minute counts.

PR #138, opened 2017-05-02
Enable support for Wayland graphics

jlaako:
Enable support for Wayland graphics through local.conf include. Since
this changes DISTRO_FEATURES checked by multiple recipes, it cannot
be static distro configuration.


mythi:
Let's pull this via #139

PR #139, opened 2017-05-02
X11

jlaako:
Note, this PR depends on PR #138 !


mythi:
@jlaako These look good to me but I feel the commit msg headers are a bit misleading because we're not enabling anything by default. Perhaps reword both commits saying 'add .inc to help enable X11/Wayland' or something like that.

while we're at it, perhaps rebase. I'll close #138 and let's take both Wayland/X11 in this PR.

jlaako:
Rebased & reworded

mythi:
@jlaako thanks!

PR #141, opened 2017-05-03
CI: refactor creation of auto.conf, use common functions

okartau:
This enables auto.conf creation from multiple scripts


okartau:
still needs added use of these functions from pre/post test scripts, currently I just kept same functionality with pieces detached into separate file, so not ready yet. I keep test label until all done.

PR #143, opened 2017-05-05
meta-iotqa: Clean up and enablement of single-node Bluetooth test cases

jairglez:
This PR cleans up and includes single-node Bluetooth test cases in the refkit-image-common manifest (this is tracked on Yocto bug #11255). It also adds bluetoothctl back to development images as some of the enabled test cases make use of this tool.

This PR replaces https://github.com/intel/intel-iot-refkit/pull/116

### DIFF
--- a/meta-iotqa/conf/test/minnowboardturbot.mask
+++ b/meta-iotqa/conf/test/minnowboardturbot.mask
@@ -1,2 +1,4 @@
 # This file contains tests that can't be run on Minnowboard Turbot
 # /docker/tester-exec.sh will use this file to remove tests from a manifest
+oeqa.runtime.connectivity.bluetooth.bt_command
+oeqa.runtime.connectivity.bluetooth.bt_6lowpan
--- a/meta-iotqa/conf/test/qemu.mask
+++ b/meta-iotqa/conf/test/qemu.mask
@@ -1,6 +1,8 @@
 # This file contains tests that can't be run on QEMU
 # /docker/tester-exec.sh will use this file to remove tests from a manifest
 oeqa.runtime.connectivity.bluetooth.btcheck
+oeqa.runtime.connectivity.bluetooth.bt_command
+oeqa.runtime.connectivity.bluetooth.bt_6lowpan
 oeqa.runtime.connectivity.wifi.wifi_connect
 oeqa.runtime.peripherals.mraa.mraa_gpio
 oeqa.runtime.multimedia.audio.alsa
--- a/meta-iotqa/conf/test/refkit-image-common.manifest
+++ b/meta-iotqa/conf/test/refkit-image-common.manifest
@@ -3,6 +3,8 @@ oeqa.runtime.core.baseos
 oeqa.runtime.connectivity.services.ssh
 oeqa.runtime.connectivity.services.managerdaemon
 oeqa.runtime.connectivity.bluetooth.btcheck
+oeqa.runtime.connectivity.bluetooth.bt_command
+oeqa.runtime.connectivity.bluetooth.bt_6lowpan
 oeqa.runtime.connectivity.wifi.wifi_connect
 oeqa.runtime.programming.python.apprt_python
 oeqa.runtime.multimedia.audio.alsa
--- a/meta-iotqa/lib/oeqa/runtime/connectivity/bluetooth/bluetooth.py
+++ b/meta-iotqa/lib/oeqa/runtime/connectivity/bluetooth/bluetooth.py
@@ -1,13 +1,11 @@
 import time
 import os
-import string
 from oeqa.utils.helper import shell_cmd_timeout
 
+
 class BTFunction(object):
-    """
-    @class BTFunction
-    """
     log = ""
+
     def __init__(self, target):
         self.target = target
         # un-block software rfkill lock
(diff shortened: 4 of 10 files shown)


meta-refkit/conf/distro/include/refkit-development.inc
```diff
@@ -33,3 +33,4 @@ REFKIT_IMAGE_BUILD_MODE_SELECTED = "1"
 
 REFKIT_IMAGE_COMPUTERVISION_EXTRA_FEATURES_append = " computervision-test"
 REFKIT_IMAGE_FEATURES_COMMON_append = " common-test"
+BAD_RECOMMENDATIONS_remove = "bluez5-client"
```
mythi: both are not needed. either this or the packagegroup change

mythi:
@jairglez thanks. did the tests work locally for you? They are failing now.

meta-refkit/conf/distro/include/refkit-development.inc
jairglez: When I only add the packagegroup entry, the build fails with a message indicating there is no provider for bluez5-client. I had to add the BAD_RECOMMENDATIONS_remove entry for the build to work. I had left the packagegroup entry to indicate the dependency for bluetoothctl in common-test, but I´ll test again without it.

jairglez:
The tests worked for a 570x, which it's the device I have available, I'm waiting for a Bluetooth card to be able to test over Turbot. For qemu, the tests definitely need to be masked out thought.

meta-refkit/conf/distro/include/refkit-development.inc
mythi: Something is not right because according to ```BAD_RECOMMENDATIONS``` is must work:

```
It is important to realize that if you choose to not install packages using this variable and some other packages are dependent on them (i.e. listed in a recipe's RDEPENDS variable), the OpenEmbedded build system ignores your request and will install the packages to avoid dependency errors.
```

mythi: There's a bug open: https://bugzilla.yoctoproject.org/show_bug.cgi?id=11427

jairglez:
Test this please

meta-refkit/conf/distro/include/refkit-development.inc
jairglez: I rebased the commit and amended it to only include the BAD_RECOMMENDATIONS_remove entry.

mythi:
@jairglez are you looking to finish this?

jairglez:
@mythi I'm still looking to finish it, the problem is in my tests with an USB Bluetooth adapter, the Minnowboard tests work correctly, but on the CI, they fail on 'hciconfig hci0 reset', however the logs don't provide any info besides that.
Long ago, I requested Bluetooth modules for the Silverjaw Lures to be able to test it, however it seems the PO had a problem because I have not received them yet. I will contact the provider again to try to confirm the order is still open and remove any blocker. Meanwhile, could you please provide me with some insight about what could be happening? I could probably remove the reset command from the tests, but I would like to know the exact cause of the failure first.

jairglez:
Thanks @mythi, I wasn't aware I could use that option. I'll execute some debugging tests then.

mythi:
@jairglez that's very common option when something works locally but not in CI setup

jairglez:
@mythi, with the latest changes to the refkit-layers what is now the best place to set up BAD_RECOMMENDATIONS_remove = "bluez5-client" for development images? Is it meta-refkit-core/classes/refkit-image.bbclass a good selection?

mythi:
@jairglez refkit-image.bbclass is OK but ```BAD_RECOMMENDATIONS_remove``` now needs to be made conditional to check ```IMAGE_MODE = 'development'``` (see examples in that bbclass)

jairglez:
Test this please

ipuustin:
You need to rebase this PR due to file conflicts.

jairglez:
Test this please

jairglez:
Test this please

jairglez:
Test this please

jairglez:
@mythi, @ipuustin: According to the test results, the Bluetooth module doesn't seem to be working at all on the Minnowboard Turbot. I consider we should file a bug for this case, and to merge this patchset and start testing the Joule at least, we should mask these tests out for Minnowboard Turbot while it is fixed. @epyeoh, do you agree?

jairglez:
@mythi, @ipuustin, @epyeoh: I finally received the same Wireless pcie adapter that it's used on the CI (Intel Dual Band Wireless-AC 7260 - 7260HMW), and I was able to test the issue further. Basically, the wireless adapter stops working after the command 'connmanctl disable bluetooth' is issued, while on dmesg, the following message is shown: "[  421.158820] Bluetooth: hci0: turning off Intel device LED failed (-61)". After that, the hci0 interface stops responding and any attempt to bring the device up times out until the device is rebooted.
Additionally, I found a similar occurrence on Red Hat: https://bugzilla.redhat.com/show_bug.cgi?id=1290011

jairglez:
I've created bug https://bugzilla.yoctoproject.org/show_bug.cgi?id=11714 detailing the previous issue. I also rebased the code again and added the new Bluetooth test cases to the mask file for MinnowBoard Turbot.

mythi:
@jairglez one thing you can easily test/try still if this is a kernel problem. please try setting ```PREFERRED_PROVIDER_virtual/kernel ?= "linux-yocto"```

jairglez:
@mythi I used the configuration you mentioned to create a custom refkit-common-image build, but the same issue occurs on it.

jairglez:
Test this please

pohly:
@jairglez is this now considered ready for merging? At least the test results are clean.

jairglez:
@pohly I would consider it ready for merging. As I previously mentioned, I had to skip testing the MinnowBoard because of bug https://bugzilla.yoctoproject.org/show_bug.cgi?id=11714 affecting the current Bluetooth device used on RefKit's CI setup. The tests can be enabled back once the issue is solved in some way, maybe by using a different device on the CI setup (I tested it successfully with Bluetooth USB adapter TRENDnet TBW-107UB), or by correcting the issue, which I suspect is driver or firmware related.

PR #144, opened 2017-05-05
Add input group because systemd is anyway creating it.

juimonen:


mythi:
@juimonen if it's being created, why is this change needed?

juimonen:
same as with audio group. it is created in first boot by systemd so you can add users to it in bitbaking.

juimonen:
so you _can't_ add users in bitbake

mythi:
@juimonen can you write that in the commit message?

mythi:
Let's track this part of #145

PR #147, opened 2017-05-08
refkit layers

pohly:
The intention is to make it simpler for other projects to use content from
intel-iot-refkit without having to use the entire distro. This first commit
is good enough to keep building the refkit distro and just moves files
around without modifying them. Further work is needed to make the
individual pieces also functional stand-alone.


okartau:
CI run (job #954) keeps doing in a loop:
swtpm_setup.sh─┬─sleep 0.1

seems like a case of endless sleep-loop we have had before in early selftest-added times

pohly:
@mythi, @RP: one question I have is around best practices for deciding between using vs. not using .bbappends. In some cases, a .bbappend has to be used (defining functions or adding files). But when merely changing variables like PACKAGECONFIG, the change can also be in a .inc file with a pn override.

We've not been consistent in refkit regarding that:
```
$ cat meta-refkit-core/recipes-multimedia/pulseaudio/pulseaudio_%.bbappend
DEPENDS_remove_refkit-config = "gconf"
EXTRA_OECONF_append_refkit-config = " --disable-gconf"

$ grep PACKAGECONFIG_.*pn meta-refkit-core/conf/distro/include/refkit-config.inc
PACKAGECONFIG_pn-python3-pygobject ?= ""
PACKAGECONFIG_remove_pn-gstreamer1.0-plugins-base = "pango"
PACKAGECONFIG_remove_pn-pulseaudio = "avahi"
PACKAGECONFIG_append_pn-ovmf_refkit-config = " secureboot"
PACKAGECONFIG_append_pn-opencv_refkit-config = " opencl"
PACKAGECONFIG_pn-gstreamer1.0-plugins-bad_refkit-config ?= ""
```

Any preference from you on this? I'm undecided myself.

rpurdie:
I think for me it depends what the reason for the change is. We group some things together as .inc files at the distro level if they make sense as a logically grouped set of changes using pn- overrides. I can see cases where something might not fit those includes/groups in which case a bbapend may be more appropriate.

mythi:
@pohly I also think that includes are preferred in cases where logical grouping (PACKAGECONFIG, SECURITY_FLAGS, etc.) is possible.  Seems like we all agree the principles. I "learnt" to use the PACKAGECONFIG pn-override late so there might be some inconsistencies.

FWIW the pulseaudio .bbappend should be moved to a patch to add a PACKAGECONFIG entry in the  .bb and then we can do the right config in refkit-config.inc.

pohly:
@mythi I think we agree on the basics, but let's be specific. So we do PACKAGECONFIG changes in the .inc  file, security flags in refkit_security_flags.inc, and everything else in .bbappends?

tripzero:
I pulled pohly/refkit-layers branch and added meta-refkit-core to my bblayers.  Builds fine.  +1

tripzero:
The dependency of image-dsk and refkit-image bbclasses is incestuous.  image-dsk requires REFKIT_VFAT_MB to be set otherwise you get a json parse error if you don't use refkit-image.bbclass.

pohly:
On Wed, 2017-05-17 at 10:20 -0700, Kevron Rees wrote:
> The dependency of image-dsk and refkit-image bbclasses is incestuous.
> image-dsk requires REFKIT_VFAT_MB to be set otherwise you get a json
> parse error if you don't use refkit-image.bbclass.

image-dsk was meant to be removed. @mythi, should we keep it as a
stand-alone class and ensure that it activates the wic image format with
UEFI combo app?

@tripzero is using refkit-image.bbclass an option? Probably not, it
brings in lots of other functionality, too.

mythi:
On Thu, May 18, 2017 at 9:08 AM, Patrick Ohly <notifications@github.com>
wrote:

> On Wed, 2017-05-17 at 10:20 -0700, Kevron Rees wrote:
> > The dependency of image-dsk and refkit-image bbclasses is incestuous.
> > image-dsk requires REFKIT_VFAT_MB to be set otherwise you get a json
> > parse error if you don't use refkit-image.bbclass.
>
> image-dsk was meant to be removed. @mythi, should we keep it as a
> stand-alone class and ensure that it activates the wic image format with
> UEFI combo app?
>
>
I'm afraid we might need a deprecation period for it. Unfortunately it does
not seem to work as a stand-alone class right now because of the
emmc-partitions-data/DSK_IMAGE_LAYOUT dependency. An easy fix
would be to skip emmc-partitions-data generation if DSK_IMAGE_LAYOUT is
empty. The stand-alone users would then set DSK_IMAGE_LAYOUT=""

pohly:
On Thu, 2017-05-18 at 05:32 -0700, Mikko Ylinen wrote:
> I'm afraid we might need a deprecation period for it.

What do you mean with that? Do you mean we have users outside of refkit
(who?), and those users need advance warning that the class will be
removed?

Right now, it is more like an internal class. We would have already
removed it if we had gotten around to it. But now it looks like it has
value on its own, if we fix it so that it actually works stand-alone.

mythi:
On Thu, May 18, 2017 at 3:50 PM, Patrick Ohly <notifications@github.com>
wrote:

> On Thu, 2017-05-18 at 05:32 -0700, Mikko Ylinen wrote:
> > I'm afraid we might need a deprecation period for it.
>
> What do you mean with that? Do you mean we have users outside of refkit
> (who?), and those users need advance warning that the class will be
> removed?
>
>
ref-os-iot uses the ostro-os-xt installer script that depends on
emmc-partitions-data info. I don't want to drag that in the new
implementation because the installer does not need it anymore. If we get
ref-os-iot to use image-installer then no deprecation is needed.

pohly:
Rebased, fixed, squashed, all my TODOs addressed -> this could be merged now.

pohly:
test this please

@mythi, @ipuustin: please review

pohly:
Note that we also agreed to standardize on putting PACKAGECONFIG into the .inc file. Cleaning that up isn't done yet and also is kind of orthogonal to the purpose of the PR, but I can add it.

pohly:
test this please

pohly:
Rebased once more, fixed refkit_image_muted, added build test of refkit-image-computervision. The latter one is potentially slow - let's see how slow. I also don't know yet whether it really passes (local build still running).

meta-refkit-core/recipes-devtools/python/python3_%.bbappend
```diff
@@ -0,0 +1 @@
+require python-readline.inc
```
ipuustin: Well, it's not only readline which is removed... also gdbm and db are not compiled in.

meta-refkit-core/conf/distro/include/refkit-config.inc
```diff
@@ -33,33 +122,22 @@ PACKAGECONFIG_pn-python3-pygobject_refkit-config = ""
 # pango, a graphical text rendering library, is not needed by us.
 PACKAGECONFIG_remove_pn-gstreamer1.0-plugins-base_refkit-config = "pango"
 
-# remove readline support from Bluez to reduce GPLv3 dependencies
-BAD_RECOMMENDATIONS_append_refkit-config = " bluez5-client"
```
ipuustin: This appears not to be recreated anywhere.

ipuustin:
Is the idea here that meta-refkit-core only depends on openembedded-core packages?

ipuustin:
It's possible that splitting meta-refkit content into a number of profile-specific layers has issues. For example, computer vision and robotics profiles will likely share a number of recipes and probably have a circular dependence between themselves.

Also, I don't see a meta-refkit-common layer. It the intention to map refkit-core to refkit-common? Would it be better to create a meta-refkit-common which would include shared recipes which could depend on layers other than openembedded-core? For example, beignet (OpenCL library) could be part of that.

pohly:
On Fri, 2017-05-19 at 06:32 -0700, Ismo Puustinen wrote:
> Is the idea here that meta-refkit-core only depends on
> openembedded-core packages?

Yes. See the introduction.rst.

pohly:
On Fri, 2017-05-19 at 06:41 -0700, Ismo Puustinen wrote:
> It's possible that splitting meta-refkit content into a number of
> profile-specific layers has issues. For example, computer vision and
> robotics profiles will likely share a number of recipes

Then those shared packages go into meta-refkit-core. My previous reply
perhaps was a bit too brief: meta-refkit-core must not have *hard*
dependencies on non-OE-core recipes, but *soft* dependencies are
allowed.

See the HAVE_* mechanism in refkit-core/layer.conf and the corresponding
checks in refkit-core for how that works.

>  and probably have a circular dependence between themselves.

Things that have a circular dependency need to be in the same layer.

> Also, I don't see a meta-refkit-common layer.

meta-refkit-core is that common layer.

pohly:
On Fri, 2017-05-19 at 06:26 -0700, Ismo Puustinen wrote:
> -# remove readline support from Bluez to reduce GPLv3 dependencies
> -BAD_RECOMMENDATIONS_append_refkit-config = " bluez5-client"
> 
> This appears not to be recreated anywhere.

Well-spotted, thanks for the thorough review. It was meant to be in the
"global" section, but got lost in the latest reshuffling. I'll fix it.

meta-refkit-core/recipes-devtools/python/python3_%.bbappend
pohly: Should I just rename it to refkit-python.inc?

pohly:
test this please

Let's retest with a more complete sstate cache...

pohly:
retest this please

pohly:
retest this please

test this please

okartau:
test this please

pohly:
test this please

okartau:
test this please

pohly:
test this please

KVM was not active in the CI, unrelated to this PR.

okartau:
test this please

pohly:
test this please

Squashed some fixes into "refkit: replace development/production .inc files with per-image IMAGE_MODE"

* plain password install test was skipped because the password was not set globally anymore
* ``inherit image-mode-variants`` had been inserted in the middle of some related chunks of settings, now it is above those

Added one new commit: c0ad16ab72f "refkit-image.bbclass: enable emergency shell in development mode"

pohly:
test this please

Previous build failed with a temporary download error:
unable to access 'https://github.com/intel-iot-devkit/upm.git/': Failed to connect to proxy.jf.intel.com port 911: Connection refused

It should have used the DLDIR from the main build, not sure why it wasn't cached. Something to investigate...

pohly:
retest this please

mythi:
test this please

pohly:
retest this please

PR #149, opened 2017-05-08
refkit-init-build-env: wrap it around oe-init-build-env

mythi:
oe-init-build-env allows to use a wrapper script that just
sets the necessary (custom) values to operate with.

Since oe-init-build-env does most of the init work, it's not needed to
maintain the same in refkit-init-build-env. Therefore, simply make
refkit-init-build-env a wrapper around oe-init-build-env. This makes
refkit-init-build-env simple.

Furthermore, this change fixes a problem where it was not possible to source
the build-env script outside the meta-data dir (intel-iot-refkit). With the
fix also it's possible to have the build directory separate from the
meta-data dir.

Fixes: [YOCTO #11402]


okartau:
test this please, seems worker error caused by recent reconfig

refkit-init-build-env
```diff
@@ -30,50 +30,23 @@ if [ -n "$BASH_SOURCE" ]; then
...
-    THIS_SCRIPT_DIR=$(dirname "$THIS_SCRIPT")
-    THIS_SCRIPT_DIR=$(readlink -f "$THIS_SCRIPT_DIR")
-    if [ -f "$THIS_SCRIPT_DIR/.templateconf" ]; then
-        source $THIS_SCRIPT_DIR/.templateconf
-    elif [ -d "$THIS_SCRIPT_DIR/.template" ]; then
-        TEMPLATECONF="$THIS_SCRIPT_DIR/.template"
-    fi
+if [ -z "$OEROOT" ]; then
+    OEROOT="$THIS_SCRIPT_DIR"/openembedded-core
 fi
```
pohly: This looks a lot saner than before.

I just wonder whether OEROOT and BITBAKEDIR are allowed to be set before sourcing refkit-init-build-env. The "if" check seems to imply that this is allowed. When used in this way, is the current .templateconf going to work? It's falls back to a relative path that won't be correct.

It fails for me, as expected:

    $ export OEROOT=/fast/work/openembedded-core
    $ . /fast/work/intel-iot-refkit/refkit-init-build-env /tmp/test
    Error: TEMPLATECONF value points to nonexistent directory '../meta-refkit/conf'

pohly: This patch on top of your tree fixes the problem:

```
diff --git a/.templateconf b/.templateconf
deleted file mode 100644
index ef125d2642..0000000000
--- a/.templateconf
+++ /dev/null
@@ -1,2 +0,0 @@
-# Template settings (relative to OEROOT)
-TEMPLATECONF=${TEMPLATECONF:-../meta-refkit/conf}
diff --git a/refkit-init-build-env b/refkit-init-build-env
index e76c26fe5c..a9a8d685f8 100755
--- a/refkit-init-build-env
+++ b/refkit-init-build-env
@@ -44,7 +44,8 @@ if [ -z "$BITBAKEDIR" ]; then
     BITBAKEDIR="$THIS_SCRIPT_DIR"/bitbake
 fi
 
-. "$THIS_SCRIPT_DIR"/.templateconf
+# Template settings (relative to OEROOT)
+TEMPLATECONF=${TEMPLATECONF:-"$THIS_SCRIPT_DIR"/meta-refkit/conf}
 
 unset THIS_SCRIPT
 unset THIS_SCRIPT_DIR
```

Note that the separate .templateconf file doesn't seem to be needed at all, so I'm removing it.

I also wonder whether should have a oe-init-build-env -> refkit-init-build-env symlink or even rename our file entirely. There's no reason for the script to be named differently, it's just another learning step for people trying to get started.

mythi:
On Thu, May 11, 2017 at 6:15 PM, Patrick Ohly wrote:

> *@pohly* commented on this pull request.
> ------------------------------
>
> In refkit-init-build-env
> <https://github.com/intel/intel-iot-refkit/pull/149#discussion_r116018377>
> :
>
> >
> -if [ -z "$TEMPLATECONF" ]; then
> -    THIS_SCRIPT_DIR=$(dirname "$THIS_SCRIPT")
> -    THIS_SCRIPT_DIR=$(readlink -f "$THIS_SCRIPT_DIR")
> -    if [ -f "$THIS_SCRIPT_DIR/.templateconf" ]; then
> -        source $THIS_SCRIPT_DIR/.templateconf
> -    elif [ -d "$THIS_SCRIPT_DIR/.template" ]; then
> -        TEMPLATECONF="$THIS_SCRIPT_DIR/.template"
> -    fi
> +if [ -z "$OEROOT" ]; then
> +    OEROOT="$THIS_SCRIPT_DIR"/openembedded-core
>  fi
>
> This patch on top of your tree fixes the problem:
>

Thanks for checking this! Looks good.

> I also wonder whether should have a oe-init-build-env ->
> refkit-init-build-env symlink or even rename our file entirely. There's no
> reason for the script to be named differently, it's just another learning
> step for people trying to get started.
>
>
>  Then it would be oe-init-build-env calling another oe-init-build-env
(OTOH it'd be mostly hidden in the script). But the proposal makes sense.
We could provide the old name as a symlink for the time being, how about
that?

pohly:
On Thu, 2017-05-11 at 09:54 -0700, Mikko Ylinen wrote:
> Then it would be oe-init-build-env calling another oe-init-build-env
> (OTOH it'd be mostly hidden in the script). But the proposal makes
> sense.

It walks, talks and behaves like oe-init-build-env, so we might as well
call it that, even if it is implemented differently.

> We could provide the old name as a symlink for the time being, how
> about that?

Sounds good to me.

mythi:
v2: we need to force OEROOT to our repo because oe-setup-builddir writes layerpaths relative to OEROOT. Renamed the script.

PR #151, opened 2017-05-10
meta-iotqa: Use fixed dir name inside testsuite tarball and improve readme

skuusela:
Depending on if `do_test_iot_export` task is done with or without docker the directory name inside created testsuite tarball will differ. So just use specific name with it ("iottest"). Also improve meta-iotqa/README.MD.


skuusela:
Changed one colon to a dot.

PR #152, opened 2017-05-10
Add a selftest for non-GPLv3 computer vision image

ipuustin:
Create a test image and test it with a selftest.


meta-iotqa/lib/oeqa/selftest/computervision-no-gplv3.py
```diff
@@ -0,0 +1,423 @@
...
+class GplV3Test(oeSelfTest):
+    """GPLv3 licensing test class."""
+
+    test_image = 'refkit-image-computervision-nogplv3'
+
+    @classmethod
+    def setUpClass(cls):
+        # Create the test image. We don't know which packages end up to
+        # the image without building it, because manifes isn't available
+        # with just "bitbake -c rootfs".
```
mythi: did you check ```tmp-glibc/deploy/licenses/<image>-<machine>-<build>/*``` That seems to be available after ```-c rootfs```.

meta-iotqa/lib/oeqa/selftest/computervision-no-gplv3.py
```diff
@@ -0,0 +1,423 @@
...
+        imagename = bb_vars["IMAGE_LINK_NAME"]
+        basedir = bb_vars["META_REFKIT_BASE"]
+        manifest = os.path.join(deploydir, imagename + ".manifest")
+        whitelist = os.path.join(basedir, "../meta-iotqa/scripts/contrib/licensetree-whitelist.txt")
+
+        self.assertTrue(os.path.islink(manifest), msg="No manifest file created for image. It should have been created in %s" % manifest)
+
+        # Whitelist contains the runtime dependencies which do not
+        # propagate licenses. For example when a component uses a D-Bus
+        # API or execs another process.
```
mythi: Just curious. why is the whitelist so long? Any guidance how the list should be kept up-to-date?

meta-iotqa/scripts/contrib/licensetree-whitelist.txt
```diff
@@ -0,0 +1,35 @@
...
+ofono
+bluez5
+connman
+util-linux
+udev
+libgcc-initial
+linux-libc-headers
+python-core
+python3-core
+kernel-4.9.20-yocto-standard
```
mythi: the kernel package name is different nowadays

meta-iotqa/lib/oeqa/selftest/computervision-no-gplv3.py
```diff
@@ -0,0 +1,423 @@
...
+
+    children = findChildren(package, whiteList, [package])
+    root = LicenseNode(licenses, children, package)
+
+    return root
+
+class GplV3Test(oeSelfTest):
+    """GPLv3 licensing test class."""
+
+    test_image = 'refkit-image-computervision-nogplv3'
```
mythi: Can you change the structure such that it's easy to add new a test per image. Also, is it not possible to set what gets installed in the test image by using ```write_config``` or what you can pass to ```bitbake()```

meta-iotqa/scripts/contrib/licensetree-whitelist.txt
pohly: The dependencies through linking are known to bitbake and are used to add the necessary RDEPENDS. Perhaps instead of using the final RDEPENDS, one could take the information about linking? Then there would be no need for a whitelist.

However, the information might not be readily available and an enhancement request might be needed.

meta-iotqa/lib/oeqa/selftest/computervision-no-gplv3.py
pohly: Long-term I envision per-image "production" vs. "development" modes, instead of the rather limited refkit-production.inc vs. refkit-development.inc. Then refkit-image-computervision could just be re-generated in the production mode by flipping a single setting in the local build configuration of the test.

Regarding adding more tests: I think the generic license checking code should be in a meta-refkit/lib/ Python module. Then it becomes easier to reuse. Some automated tests for it would also be useful, to ensure that it really catches the errors that it is meant to catch.

meta-iotqa/lib/oeqa/selftest/computervision-no-gplv3.py
ipuustin: Indeed the manifest is there. I'll fix this for the next iteration.

meta-iotqa/lib/oeqa/selftest/computervision-no-gplv3.py
ipuustin: The whitelist might have some components which are not present (I tested this on Poky too). Another way to approach this would be to have a "blacklist" of library components which can cause license propagation. I checked the `SECTION` variable but it appears to be unrealiable: few recipes do package-level assignment to sections, some libraries are assigned to `devel` or `utils` instead of `libs`, and so on.

Since adding more licenses to the runtime dependency tree should only add constraints, so having a component forgotten from the whitelist would only result in possible false positives, meaning that the test will fail when it should have succeeded.

meta-iotqa/scripts/contrib/licensetree-whitelist.txt
ipuustin: Using the linking information would help a lot, yes. However, it doesn't solve the (possible) issues with non-C components, such as node.js or Python libraries. Maybe those recipes could then be blacklisted based on the inherited classes or something?

meta-iotqa/lib/oeqa/selftest/computervision-no-gplv3.py
ipuustin: @pohly: Who tests the tests? :-) I could try to refactor this to `lib` and make it be somehow more generally useful.

ipuustin:
Rebased and refactored into a separate library. I did the `-c rootfs` optimization.

ipuustin:
Rebased.

okartau:
test this please, for some reason not triggered after push

meta-iotqa/lib/oeqa/selftest/computervision-no-gplv3.py
mythi: @ipuustin I don't see this feedback was yet addressed. I'm not convinced we test specific image .bb is needed. Also, the test class could be more generic to test any image

```python
def check_licensing(self, image)
    ...

def test_computervision_image(self)
    # write image mode config
    bitbake(image)
    check_licensing(image)

def test_gateway_image(self)
    # write image mode config
    bitbake(image)
    check_licensing(image)

```

ipuustin: @mythi Ok, now I think I understand what you mean. I'm not sure if we can easily change the image mode (development -> production) in the tests, even though this would definitely be closer to the real world than just changing development image content. I'll investigate this deeper.

pohly:
On Thu, 2017-05-18 at 07:02 -0700, Ismo Puustinen wrote:
> I'm not sure if we can easily change the image mode (development ->
> production) in the tests, even though this would definitely be closer
> to the real world than just changing development image content. I'll
> investigate this deeper.

With PR #147, that's (conceptually) as easy as:

self.append_config('IMAGE_MODE=production')

However, in practice production images also need custom dm-verity keys,
so in addition one also has to add:

self.append_config('REFKIT_DMVERITY_PRIVATE_KEY ="${META_REFKIT_CORE_BASE}/files/dm-verity/private.pem"')
self.append_config('REFKIT_DMVERITY_PASSWORD = "pass:refkit"')

We could add a "production-content" IMAGE_MODE which produces images
with the same content as production images, but uses the unsafe
configuration from the "development" (auto-login, known keys). Not sure
whether it's worth it, though.

ipuustin:
Rebased on top of #147. Implemented the configuration file change that @mythi asked with the mechanism @pohly suggested. Remvoed `refkit-image-computervision-nogplv3` image. Cleaned up the whitelist.

Merging of this PR should be delayed until #147 is either merged or closed. If it's merged, this can be rebased on top of master.

mythi:
@ipuustin looks good now!

ipuustin:
retest this please

ipuustin:
Rebased.

ipuustin:
retest this please

ipuustin:
test this please

PR #153, opened 2017-05-11
refkit-installer-image: do not depend on emmc-partitions-data

mythi:


meta-refkit/recipes-image/images/refkit-installer-image.bb
```diff
@@ -191,29 +191,28 @@ REFKIT_INSTALLER_UEFI_COMBO () {
...
-            eval type_id="\$PART_${pnum}_TYPE"
-            eval lname="\$PART_${pnum}_NAME"
-            eval fs="\$PART_${pnum}_FS"
+        while [ "$gdisk_pnum" -le "$PART_COUNT" ]; do
+            size=$(execute sgdisk -i $gdisk_pnum "$input" | grep ^"Partition size" | cut -d : -f 2 | awk '{print $1}')
+            type_id=$(execute sgdisk -i $gdisk_pnum "$input" | grep ^"Partition GUID code" | cut -d : -f 2 | awk '{print $1}')
+            lname=$(execute sgdisk -i $gdisk_pnum "$input" | grep ^"Partition name" | cut -d : -f 2 | awk '{print $1}' | tr -d \')
 
+            if [ "$lname" = "rootfs" ]; then
+                uuid=$(execute grep PARTUUID "$input_mountpoint"/boot/EFI_internal_storage/BOOT/boot*.efi | sed -e 's/.*PARTUUID=\([a-zA-Z0-9-]*\).*/\1/')
```
pohly: This is non-obvious and should come with a comment about REMOVABLE_MEDIA_ROOTFS_PARTUUID_VALUE vs. INT_STORAGE_ROOTFS_PARTUUID_VALUE  and how we ensure here that we use the value expected by the kernel.

Also, instead of plain grep it might be better to use "strings "$input_mountpoint"/boot/EFI_internal_storage/BOOT/boot*.efi | grep PARTUUID". IMHO strings is more likely to output one string per line than grep.

mythi:
On Thu, May 11, 2017 at 4:00 PM, Patrick Ohly wrote:

> *@pohly* commented on this pull request.
> ------------------------------
>
> In meta-refkit/recipes-image/images/refkit-installer-image.bb
> <https://github.com/intel/intel-iot-refkit/pull/153#discussion_r115982939>
> :
>
> >
> +            if [ "$lname" = "rootfs" ]; then
> +                uuid=$(execute grep PARTUUID "$input_mountpoint"/boot/EFI_internal_storage/BOOT/boot*.efi | sed -e 's/.*PARTUUID=\([a-zA-Z0-9-]*\).*/\1/')
>
> This is non-obvious and should come with a comment about
> REMOVABLE_MEDIA_ROOTFS_PARTUUID_VALUE vs. INT_STORAGE_ROOTFS_PARTUUID_VALUE
> and how we ensure here that we use the value expected by the kernel.
>

I'll add a comment. It will be MUCH easier to explain the case here. With
the emmc-partitions-data it's not obvious at all how the rootfs PARTUUID
gets set.


> Also, instead of plain grep it might be better to use "strings
> "$input_mountpoint"/boot/EFI_internal_storage/BOOT/boot*.efi | grep
> PARTUUID". IMHO strings is more likely to output one string per line than
> grep.
>
>
> I wasn't familiar with strings. I'll change it.

PR #140, opened 2017-05-02
oeqa/runtime: Restructuring runtime test suite

pacopedraza:
Reestructuration of all the test suite for runtime
Created the following directories:
bsp, conformance, connectivity, core, multimedia, peripherals, programming

Signed-off-by: Francisco Pedraza <francisco.j.pedraza.gonzalez@intel.com>

PR #155, opened 2017-05-12
meta-iotqa: Reorganized test modules

ibarraJhumberto:
Reorganize test modules in a way that makes more sense. Tests are
now classified by component. Also, remove unneeded 'comm' prefix
in some modules.

Update paths to modules in manifests. Also, sanity tests are better
grouped as part of manifest than module location, so moved sanity
test modules to proper location.

Signed-off-by: Humberto Ibarra <humberto.ibarra.lopez@intel.com>


ibarraJhumberto:
This pull request replaces #140 with the changes proposed there. Made everything in a single commit to avoid breaking the manifest between commits.

skuusela:
This way of organizing the tests indeed looks better than the current way.

ibarraJhumberto:
Rebased fork to latest commit.

PR #158, opened 2017-05-17
Add Intel(R) Math Kernel Library for Deep Neural Networks.

ipuustin:
Add a new package. It can be used as a building block for higher-level DNN frameworks.


ipuustin:
Rebased.

PR #159, opened 2017-05-17
submodule updates

mythi:


ipuustin:
There's now a bug abouit the boost ublas build failure: https://github.com/boostorg/ublas/issues/47. Seems that a patch already exists.

PR #162, opened 2017-05-23
Clean up beignet recipe

ipuustin:
Enable parallel build for beignet by fixing cmake dependencies. Also remove host contamination possibilities by using Python from sysroot instead of Python from build host. This should also help with the incompatible host clang version problem. Also update beignet to use the latest git HEAD.


ipuustin:
The cmake patch got accepted upstream, so just update the SRCREV.

ipuustin:
Rebased.

mythi:
test this please

PR #168, opened 2017-05-30
sbsigntool: stop using gitsm fetcher

mythi:
gitsm fetcher does not work well with download cache
(submodules never get to the local source mirror) and each
builds ends up cloning ccan.

Move to use git fetcher (with https protocol) for both sbsigntool
and ccan independently to speed up fetching and to get the
mirroring benefits.


ipuustin:
I was thinking the same. We are hacking around an (undocumented?) bug in Bitbake. Please at least file a bug for BB. We can revert this patch when BB gets fixed.

mythi:
FWIW, the limitations *are* documented in bitbake user manual:

```
Notes and Warning

The Git Submodules fetcher is not a complete fetcher implementation. The fetcher has known issues where it does not use the normal source mirroring infrastructure properly. 
```

In this particular case I don't see a big risk as the submodule has never been updated in the repo. I do understand the problem still and I'm fine submitting the bug (while submitting another issue I discovered when working on this change)

mythi:
v2: rebased + added a note (with a bug ID) about gitsm limitations

PR #172, opened 2017-06-05
python: runtime and packaging changes

mythi:
After dropping gdbm and db from DEPENDS, we also need to
make sure the corresponding ${PN}-* packages don't get
added in PACKAGES/PROVIDES/RDEPENDS.

Without this, the package managers run into dependency issues
when installing, e.g., python(3)-modules that depend on
existent packages.

While we're at it, it's good time to move python-runtime image
feature to use Python3.


ipuustin:
See my version at https://github.com/intel/intel-iot-refkit/pull/113/commits/778ad99d9654c631654d544e049ce5f0970ae411. Was there a `python3-db` package for Python 3?

mythi:
@ipuustin oops I had missed your patch. python(3)-db is there for both versions.

PR #173, opened 2017-06-06
Move IMAGE_DSK_BASE variable to correct place

jlaako:


mythi:
On Tue, Jun 6, 2017 at 8:47 PM, Patrick Ohly wrote:

> IMHO instead of fixing IMAGE_DSK_BASE and the "dsk" IMAGE_FSTYPES, we
> should remove the entire dead code.
>

I agree. I had intentionally postponed the removal but if it's not working
anymore with the new layer setup, it's time to do it.

FWIW, I also noticed today we can drop image_types_extra.bbclass too...

jlaako:
Well, it does work with this fix. I'm testing it daily...

mythi:
@jlaako please prepare to move to the refkit default image format (.wic) as .dsk is EOL.

mythi:
obsoleted by #187

PR #182, opened 2017-06-13
Update meta-security-isafw to remove cve-check-tool recipe.

ipuustin:
There appears to be an issue with cve-check-tool from meta-security-isafw. See if oe-core version of the tool works better.


ipuustin:
An alternative to this patch should be merging https://github.com/01org/meta-security-isafw/pull/84 and then updating the layer.

ipuustin:
It seems that `isafw.bbclass` keeps working after this change. Also the cve check tool results are available again.

mythi:
@ipuustin thanks!

ipuustin:
Rebased.

PR #187, opened 2017-06-14
images: drop .dsk, clean-ups, and CI optimizations

mythi:


mythi:
I don't think the argument is totally misleading: the conversion is redundant and for CI-purposes file download integrity check is sufficient. I agree the wording could be improved though.

Anyway, thanks for the review.

This obsoletes #173

PR #188, opened 2017-06-14
meta-flatpak: a flatpak layer with basic flatpak support.

klihub:
This patchset adds the usrmerge patchset from @avalluri, a few minor fixes in top of that and a new meta-flatpak layer. meta-flatpak adds support for building flatpak-enabled runtime and SDK images as well as for generating and populating a corresponding flatpak ostree repository with related branches. These are generally referred to as the BasePlatform and BaseSdk runtimes in the flatpak nomenclature. The BaseSdk runtime can then be used to generate (IOW compile/cross-compile, package and publish) applications as flatpaks for a flatpak enabled runtime image.

This patchset contains basic flatpak support, including

1) flatpak and its dependencies (including ostree)
2) a simple systemd service to set up an emulated runtime for the image using mostly bind-mounts
3) a helper class/script for generaring signing keys (used by/for signing flatpak and ostree repositories)
4) helper classes/scripts for generating/populating a per-image primary flatpak repo and optionally 
exporting it to clients (and development hosts) for consumption using flatpak

There is another patchset landing on top of this which adds

1) a proof-of-concept daemon which can be used as
  - a systemd generator during boot to set up 'flatpak sessions' for monitored flatpak repositories
  - a helper binary to launch flatpak applications belonging to a 'flatpak session'
  - a daemon to update flatpak applications, and restart them if needed
2) optionally a set of pre-declared flatpak repositories from which devices running the image will start pulling in applications to be run within 'flatpak sessions'


meta-refkit/conf/bblayers.conf.sample
```diff
@@ -20,6 +20,7 @@ REFKIT_LAYERS = " \
   ##OEROOT##/../meta-security-isafw \
   ##OEROOT##/../meta-intel-realsense \
   ##OEROOT##/../meta-clang \
+  ##OEROOT##/../meta-flatpak \
```
mythi: This justifies ```LCONF_VERSION``` bump.

pohly:
On Fri, 2017-06-16 at 01:04 -0700, Mikko Ylinen wrote:
> @mythi commented on this pull request.
> 
> 
> ______________________________________________________________________
> 
> In meta-refkit/conf/bblayers.conf.sample:
> 
> > @@ -20,6 +20,7 @@ REFKIT_LAYERS = " \
>    ##OEROOT##/../meta-security-isafw \
>    ##OEROOT##/../meta-intel-realsense \
>    ##OEROOT##/../meta-clang \
> +  ##OEROOT##/../meta-flatpak \
> 
> This justifies LCONF_VERSION bump.

Beware that LCONF_VERSION needs to be bumped to 8 unless we get my
OE-core patch "bblayers sanity: bblayers.conf.sample specific update
code" merged first.

See
https://github.com/pohly/intel-iot-refkit/commit/880610028b3eff77a860c122ca67e225338b58de

meta-flatpak/conf/layer.conf
```diff
@@ -0,0 +1,16 @@
...
+    ${LAYERDIR}/recipes-*/*/*.bbappend"
+
+BBFILE_COLLECTIONS += "flatpak-layer"
+BBFILE_PATTERN_flatpak-layer = "^${LAYERDIR}/"
+BBFILE_PRIORITY_flatpak-layer = "6"
+
+LAYERDEPENDS_flatpak-layer = "core openembedded-layer filesystems-layer"
+
+# Set a variable for easy access to the top directory of the flatpak layer.
+FLATPAK_LAYERDIR = '${@os.path.normpath("${LAYERDIR}/../meta-flatpak")}'
```
pohly: Why not just '${@os.path.normpath("${LAYERDIR}")}' ?

klihub: Good point, fixed this one.

meta-refkit/conf/bblayers.conf.sample
klihub: @mythi, @pohly So what LCONF_VERSION should be bumped to ?
5 or 8 ?
Or does this one go to 11 ?

mythi: @klihub based on the commit message from @pohly 8 it is because everything between 4...7 is "reserved".

mythi: Hmm... even 26ee25f is not enough. For some reason the bblayer.conf incompatiblity goes unnoticed and bitbake fails with missing flatpak.inc (since I don't have meta-flatpak enabled)

meta-refkit-core/conf/distro/include/enable-refkit-config.inc
```diff
@@ -15,3 +15,6 @@ VIRTUAL-RUNTIME_initscripts = ""
 # defaults from OE-core.
 DISTRO_FEATURES_DEFAULT_remove = "${REFKIT_DEFAULT_DISTRO_FEATURES_REMOVE}"
 DISTRO_FEATURES ?= "${DISTRO_FEATURES_DEFAULT} ${DISTRO_FEATURES_LIBC} ${REFKIT_DEFAULT_DISTRO_FEATURES}"
+
+require conf/distro/include/usrmerge.inc
+require conf/distro/include/flatpak.inc
```
pohly: enable-refkit-config.inc must be usable even when meta-flatpak is not used (because meta-refkit-core and thus enable-refkit-config.inc must be usable without that layer).

This is causing the "bitbake fails with missing flatpak.inc" error.

Try "include conf/distro/include/flatpak.inc" instead.

klihub: Interesting. All the tests passed in my test build without any such failures (at least I did not stop any). I don't want to interfere now/any more with Mikkos release attempts if he's willing to take this as such, but I change these to includes and do queue them for my next PR.

klihub: Okay, changed the require to an include.

pohly:
On Fri, 2017-06-16 at 13:38 -0700, Krisztian Litkey wrote:
> Interesting. All the tests passed in my test build without any such
> failures (at least I did not stop any).

It only triggers when a developer doesn't already have meta-flatpak in
their bblayers.conf - you always had that, and thus never saw this.

mythi:
On Sat, Jun 17, 2017 at 10:21 PM, Patrick Ohly <notifications@github.com>
wrote:

> On Fri, 2017-06-16 at 13:38 -0700, Krisztian Litkey wrote:
> > Interesting. All the tests passed in my test build without any such
> > failures (at least I did not stop any).
>
> It only triggers when a developer doesn't already have meta-flatpak in
> their bblayers.conf - you always had that, and thus never saw this.
>
>
What's not quite right is that bitbake only gave the flatpak.inc parse
error but did
not check layer conf version.

pohly:
On Sun, 2017-06-18 at 22:13 -0700, Mikko Ylinen wrote:
> On Sat, Jun 17, 2017 at 10:21 PM, Patrick Ohly
> <notifications@github.com>
> wrote:
> 
> > On Fri, 2017-06-16 at 13:38 -0700, Krisztian Litkey wrote:
> > > Interesting. All the tests passed in my test build without any
> such
> > > failures (at least I did not stop any).
> >
> > It only triggers when a developer doesn't already have meta-flatpak
> in
> > their bblayers.conf - you always had that, and thus never saw this.
> >
> >
> What's not quite right is that bitbake only gave the flatpak.inc parse
> error but did not check layer conf version.

I agree, that could be improved - not sure whether that's possible,
though.

The reason is that the version check only triggers after parsing the
base configuration, but that already fails in a fatal way.

PR #193, opened 2017-06-16
oe-init-build-env: fix Bash version inconsistencies

mythi:
Some versions* of Bash do not seem to set BASH_SOURCE
correctly resulting broken THIS_SCRIPT path and the
build environment setup fails.

Simplify the if elsif else logic to default to $0 if
BASH_SOURCE is not set.

Tested on:
   * GNU bash, version 4.3.30(1)
   * GNU bash, version 4.4.11(1)
   * zsh 5.0.7

[YOCTO: #11640]

* Bash 4.3.48(1) on Ubuntu 16.04 this fails but with Bash
4.2.46(1) on CentOS 7 it works.

### DIFF
--- a/oe-init-build-env
+++ b/oe-init-build-env
@@ -27,10 +27,8 @@
 #
 if [ -n "$BASH_SOURCE" ]; then
     THIS_SCRIPT=$BASH_SOURCE
-elif [ -n "$ZSH_NAME" ]; then
-    THIS_SCRIPT=$0
 else
-    THIS_SCRIPT="$(pwd)/oe-init-build-env"
+    THIS_SCRIPT=$0
 fi
 
 THIS_SCRIPT_DIR=$(dirname "$THIS_SCRIPT")


ipuustin:
Appears to work under zsh too. @pohly is right in that this should also be upstreamed.

incandescant:
I'm not sure it does work for poky, the reason we've seen this with refkit and not poky is that we have refkit setup differently on the YP AB in that we source the oe-init-build-env from a parent directory, rather than directory of the script.

For OE and Poky builds we do:
$ . oe-init-build-env

For refkit builds we do:
$ . refkit/oe-init-build-env

The reason for doing things differently is due to the way the yocto-autobuilder sets up the directory structure on layer fetch, it assumes the top-level directory is a clone of the poky repo. We want this to also be the case in nightly-refkit as we want to be able to build refkit against the in-testing, i.e. master-next/mut/etc, branches of bitbake, OE-Core, etc. Therefore we clone the intel-iot-refkit repo into a child of the poky tree and set up bblayers.conf appropriately.

ipuustin:
@incandescant : thanks for the info! Would it be possible to test this change in YP AB before we merge it to Refkit?

incandescant:
This doesn't seem to help on ubuntu1604 (GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)):

https://autobuilder.yocto.io/builders/nightly-refkit/builds/16
https://autobuilder.yocto.io/builders/nightly-refkit/builds/16/steps/RunPreamble/logs/stdio

Updating my test script from [YOCTO: #11640]

```
joshuagl@ubuntu1604:~$ source test/test.sh 
BASH_SOURCE=test/test.sh
THIS_SCRIPT=test/test.sh
$0=-bash
joshuagl@ubuntu1604:~$ cat test/test.sh 
#!/usr/bin/env sh

if [ -n "$BASH_SOURCE" ]; then
    THIS_SCRIPT=$BASH_SOURCE
else
    THIS_SCRIPT=$0
fi

echo BASH_SOURCE=$BASH_SOURCE
echo THIS_SCRIPT=$THIS_SCRIPT
echo "\$0"=$0
```

pohly:
On Mon, 2017-06-19 at 03:22 -0700, Joshua G Lock wrote:
> I'm not sure it does work for poky, the reason we've seen this with
> refkit and not poky is that we have refkit setup differently on the YP
> AB in that we source the oe-init-build-env from a parent directory,
> rather than directory of the script.
> 
> For OE and Poky builds we do:
> $ . oe-init-build-env

In other words, the same problem probably also exists in OE-core and
Poky and it needs the same fix?

That explains it for me, so I think we should go ahead and merge this
for Refkit. IMHO there's no need to do a trial-run in the AB, I trust
that Mikko has checked that it really works. OTOH, has he tested .
<path>/oe-init-build-env? Not sure :-/

ipuustin:
@incandescant Why does it fail then? `THIS_SCRIPT` seems to have the right value in all the tests that you ran. Is `/bin/sh` *dash* in Ubuntu? It would explain why @mythi reports `BASH_SOURCE` being wrongly set in some bash versions. :-)

pohly:
On Mon, 2017-06-19 at 05:25 -0700, Joshua G Lock wrote:
> This doesn't seem to help on ubuntu1604 (GNU bash, version
> 4.3.48(1)-release (x86_64-pc-linux-gnu)):
> 
> https://autobuilder.yocto.io/builders/nightly-refkit/builds/16
> https://autobuilder.yocto.io/builders/nightly-refkit/builds/16/steps/RunPreamble/logs/stdio
>
> Updating my test script from [YOCTO: #11640]
> 
> joshuagl@ubuntu1604:~$ source test/test.sh 
> BASH_SOURCE=test/test.sh
> THIS_SCRIPT=test/test.sh
> $0=-bash

So it looks like Mikko hasn't tested the out-of-directory invocation
case after all.

Given the output above, I don't see a solution for those bash versions
which set neither BASH_SOURCE nor $0. Any bright ideas from anyone?

Joshua, can you perhaps work around this by invoking the script inside
the intel-iot-refkit directory? It has not become clear to me why you
need to do this differently for refkit and poky.

pohly:
On Mon, 2017-06-19 at 05:25 -0700, Joshua G Lock wrote:
> This doesn't seem to help on ubuntu1604 (GNU bash, version
> 4.3.48(1)-release (x86_64-pc-linux-gnu)):
> 
> https://autobuilder.yocto.io/builders/nightly-refkit/builds/16
> https://autobuilder.yocto.io/builders/nightly-refkit/builds/16/steps/RunPreamble/logs/stdio
> 
> Updating my test script from [YOCTO: #11640]
> 
> joshuagl@ubuntu1604:~$ source test/test.sh 
> BASH_SOURCE=test/test.sh
> THIS_SCRIPT=test/test.sh
> $0=-bash

Sorry, I was confused.

Joshua, the script works when using bash on ubuntu1604 as your output
shows. It fails when using dash:

$ dash
$ . test/test.sh
BASH_SOURCE=
THIS_SCRIPT=dash
$0=dash

So perhaps as a workaround on ubuntu1604 avoid running the AB wrapper
script under dash, i.e. source in a bash shell?

ipuustin:
I updated the bug report. @incandescant, we can close this PR if you also feel AB is right place for fixing the issue.

incandescant:
Apologies for the noise all, I had managed to completely forget about Ubuntu's use of dash. I will try and get this worked around in yocto-autobuilder. Feel free to close.

ipuustin:
Closing. We can reopen if needed later.

PR #196, opened 2017-06-16
meta-openembedded: update to latest HEAD.

ipuustin:
Just update meta-openembedded. This is to see if the installer selftests fail with this.


ipuustin:
Git bisect says that the installer tests fail because of this meta-openembedded commit:     `lvm2/libdevmapper: fix invoking lvcreate failed` (https://github.com/openembedded/meta-openembedded/commit/6aa1dd223a140abc9f7fb6c9e9bf7134abfde5cc).

PR #198, opened 2017-06-16
Flatpak session (a.k.a. flatpak-based app fw) support.

klihub:
This patch set adds flatpak-session support to meta-flatpak on top of the existing basic flatpak support.

This mostly consists of

1) a daemon for monitoring configured flatpak remotes, pulling in/updating published apps
2) a systemd generator to set up 'flatpak sessions' for monitored flatpak repositories during boot
3) a helper binary to launch flatpak applications belonging to a 'flatpak session'
4) optionally a set of pre-declared flatpak remotes/repositories to preconfigure the images with (IOW set up remote URL, signing public key, and associated session user)

flatpak-session support has its dedicated DISTRO_FEATURE ('flatpak-session'). There is also a corresponding flatpak-session.inc configuration fragment which is used to enable flatpak-sessions.

By default flatpak-session support is disabled.

### DIFF
--- a/meta-flatpak/recipes-flatpak/flatpak-utils/files/flatpak-utils/Makefile.am
+++ b/meta-flatpak/recipes-flatpak/flatpak-utils/files/flatpak-utils/Makefile.am
@@ -1,0 +1,6 @@
+SUBDIRS  = . src
+doc_DATA = AUTHORS ChangeLog NEWS README
+
+# cleanup
+clean-local:
+	rm -f *~
--- a/meta-flatpak/recipes-flatpak/flatpak-utils/files/flatpak-utils/src/filesystem.c
+++ b/meta-flatpak/recipes-flatpak/flatpak-utils/files/flatpak-utils/src/filesystem.c
@@ -1,0 +1,335 @@
+/*
+ * Copyright (c) 2016, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *   * Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *   * Neither the name of Intel Corporation nor the names of its contributors
+ *     may be used to endorse or promote products derived from this software
+ *     without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <errno.h>
+#include <stdarg.h>
+#include <fcntl.h>
+#include <limits.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/wait.h>
+
+#include "flatpak-session.h"
+
+
+int fsys_prepare_session(context_t *c)
+{
+    char *dir = fsys_mkpath(NULL, 0, "%s/%s.wants",
+                            c->service_dir, FPAK_SYSTEMD_TARGET);
+
+    log_info("creating service directory %s...", dir);
+
+    if (c->dry_run)
+        return 0;
+    else
+        return fsys_mkdirp(0755, dir);
+}
+
+
+char *fsys_mkpath(char *path, size_t size, const char *fmt, ...)
+{
+    static char buf[PATH_MAX];
+    va_list     ap;
+    int         n;
+
+    if (path == NULL) {
+        path = buf;
+        size = sizeof(buf);
+    }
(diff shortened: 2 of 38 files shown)


meta-flatpak/recipes-flatpak/flatpak-utils/files/flatpak-utils/src/flatpak.c
```diff
@@ -0,0 +1,1011 @@
...
+        return def;
+
+    val = g_key_file_get_value(m, sec, key, NULL);
+
+    if (val == NULL)
+        return def;
+
+    i = strtol(val, &e, 10);
+
+    if (!e && !*e)
```
ipuustin: Should be `if (e && !*e)`?

klihub: Thanks, well spotted.

meta-flatpak/recipes-flatpak/flatpak-utils/flatpak-utils_git.bb
```diff
@@ -0,0 +1,40 @@
...
+    ${systemd_unitdir}/system/flatpak-sessions.target \
+    ${systemd_unitdir}/system/flatpak-session@.service \
+    ${systemd_unitdir}/system/flatpak-update.service \
+"
+
+FILES_${PN}-dbg =+ "${base_libdir}/systemd/system-generators/.debug"
+
+SYSTEMD_PACKAGES      += "${PN}"
+SYSTEMD_SERVICE_${PN}  = "flatpak-sessions.target flatpak-update.service"
+SYSTEMD_AUTO_ENABLE    = "enable"
```
ipuustin: This is `enable` by default, no need to set it. And anyway if you set it, you might want to use `?=`.

meta-flatpak/recipes-flatpak/flatpak-utils/flatpak-utils_git.bb
```diff
@@ -0,0 +1,40 @@
...
+inherit autotools pkgconfig requires-systemd
+
+AUTO_LIBNAME_PKGS = ""
+
+S = "${WORKDIR}/git"
+
+EXTRA_OECONF += "--with-systemdunitdir=${systemd_unitdir}"
+
+# possible package configurations
+PACKAGECONFIG ??= ""
```
ipuustin: Hmm, why is this needed?

meta-flatpak/recipes-flatpak/flatpak-utils/flatpak-utils_0.0.bb
```diff
@@ -7,16 +7,14 @@ LIC_FILES_CHKSUM = "file://LICENSE-BSD;md5=f9f435c1bd3a753365e799edf375fc42"
 DEPENDS = "flatpak systemd"
 
 SRC_URI = " \
-    git://git@github.com/klihub/flatpak-utils.git;protocol=https;branch=master \
+    file://flatpak-utils \
```
ipuustin: You shouldn't patch a file which you introduced in the same PR. Just change it.

meta-flatpak/recipes-flatpak/flatpak-utils/files/flatpak-utils/Makefile.am
```diff
@@ -0,0 +1,32 @@
+SUBDIRS  = . src #orig
+doc_DATA = AUTHORS ChangeLog NEWS README
+
+# This is the only way with automake I know of to force 'check-git-hooks'
+# to be evaluated before 'all'. If there is a nicer way, I'm all ears...
+BUILT_SOURCES = install-git-hooks
```
ipuustin: You don't need to install git hooks, if you have the source distribution inside the layer. There is no corresponding `.git` directory.

meta-flatpak/recipes-flatpak/flatpak-utils/files/flatpak-utils/src/filesystem.c
```diff
@@ -0,0 +1,335 @@
...
+#include <fcntl.h>
+#include <limits.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/wait.h>
+
+#include "flatpak-session.h"
+
+
+int fsys_prepare_session(context_t *c)
```
ipuustin: You anyway need to link against glib (you are using `GError` etc. in the flatpak interfacing code). Have you checked if you could just just glib file system functions (see https://developer.gnome.org/glib/stable/glib-File-Utilities.html)? That would reduce the amount of custom code we have to maintain.

ipuustin:
A README file for flatpak-utils would be nice. There are functions which are either dead code or used from somewhere I don't see, like `fpak_app_systemd_scope`. Also the restart logic is not fully clear. Is the whole daemon restarted when any flatpak app is updated? Or are you running one daemon for each app?

meta-flatpak/recipes-flatpak/flatpak-utils/flatpak-utils_git.bb
klihub: Removed SYSTEMD_AUTO_ENABLE.

meta-flatpak/recipes-flatpak/flatpak-utils/flatpak-utils_git.bb
klihub: Removed empty PACKAGECONFIG.

meta-flatpak/recipes-flatpak/flatpak-utils/flatpak-utils_0.0.bb
klihub: Squashed commits.

meta-flatpak/recipes-flatpak/flatpak-utils/files/flatpak-utils/Makefile.am
klihub: Ripped out githooks and removed associated rule.

meta-flatpak/recipes-flatpak/flatpak-utils/files/flatpak-utils/src/filesystem.c
klihub: Sure I could have used. glib has a g-wrapper for every imaginable POSIX filesystem-related system call and then some. It was an arbitrary but deliberate choice to use glib where I have to, mostly for interfacing with libflatpak.

klihub:
I've added a very brief README and removed the unused function. 

The restart logic is that whenever anything changes (gets updated) in a session, the whole session gets restarted. The actual restart is delegated to systemd. When a restart is desired, the daemon exits with a special exit code reserved for this purpose which systemd is let know about in the service file. Systemd will then restart the daemon which will then relaunch the applications within the session. At least that's the general idea...

klihub:
Well, not quite yet. I need to still migrate some documentation over from the old repo. Also while I have added basic sanity tests, we still need to make sure they still work with latest meta-iotqa and CI infra. Once the tests work and the docs are in place, then I think this would be ready to go in.

klihub:
Unfortunately running the tests for the flatpak-enabled gateway runtime image was not enabled at all. Fixed it. We'll have to check once the latest queued build is finished. Someone with supercow powers could show mercy on the currently running futile builds using the /stop REST API...

klihub:
The test were run and have passed. I've also the added initial documentation snippets about flatpak support.

klihub:
teat this please

okartau:
test this please (there needs to be 's' in 1st word :-)

klihub:
Thanks! Oh boy, I hate touchscreens...

klihub:
@ipuustin Now, this could go in. If we don't want to enable flatpak sessions by default, we should leave [the config change commit ](https://github.com/klihub/intel-iot-refkit/commit/2c64cceff37a81deb14758d8369cda6e4fa8ea58) out.

Arthil:
Retest this please.

PR #204, opened 2017-06-22
yocto-compatible.rst: implementation documentation

pohly:
This targets developers and explains how layers meet the "Yocto
Compatible 2.0" criteria and provide optional functionality to other
distributions.

### DIFF
--- a/doc/howtos/yocto-compatible.rst
+++ b/doc/howtos/yocto-compatible.rst
@@ -1,0 +1,190 @@
+Yocto Compatible 2.0
+====================
+
+The main goal of the next revision of the Yocto compliance program,
+called "Yocto Compatible 2.0", is to increase inter-operability of
+different layers. One key aspect is that a layer must not change a
+build configuration merely because it gets added to a build. Instead,
+only some explicit build configuration change (like setting distro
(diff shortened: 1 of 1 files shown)


ipuustin:
If these are generic guidelines for configuring Yocto 2.0 -compatible layers, should this be rather submitted to oe-core? We could just refer to the documentation from here and tell about the local conventions ("we use `refkit` as the distro feature, so use `df-refkit` in variable names" etc).

pohly:
On Thu, 2017-06-22 at 02:25 -0700, Ismo Puustinen wrote:
> If these are generic guidelines for configuring Yocto 2.0 -compatible
> layers, should this be rather submitted to oe-core?

Long-term, yes. But OE documentation is undergoing some replanning
(rethinking the "reference" vs. "howto" aspects), so I wouldn't know
where to put it in the current OE documentation.

>  We could just refer to the documentation from here and tell about the
> local conventions ("we use refkit as the distro feature, so use
> df-refkit in variable names" etc).

Yes, we could do that.

samniem:
Flashing froze in CI for some reason. Retest this please.

PR #205, opened 2017-06-22
CI: fix cleanup of older build directories

okartau:
1. Cleanup of older build trees was not reliable.
Cleaning should run in early stage (using parallel stmt to avoid blocking other activity).
Previous way, running it later in parallel with on-device testing created 2 problems:
- in case of queued runs, same slot got busy with another job before cleaning thread,
  which caused dummy wait of current job by length on another, newly started job:
- in case of failure during post-build, cleanup was not run at all,
  which risked filling builder disks after repeated failures on same node.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,11 +26,9 @@ def testinfo_data = ""
 def ci_git_commit = ""
 def global_sum_log = ""
 def added_commits = ""
-def builder_node = ""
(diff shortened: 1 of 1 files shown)


okartau:
test this please

okartau:
test this please

okartau:
tezt this

okartau:
tezt this

PR #208, opened 2017-06-26
OSTree system update

pohly:
Ready for review.

### DIFF
--- a/meta-refkit-core/recipes-ostree/initramfs-framework-ostree/files/ostree
+++ b/meta-refkit-core/recipes-ostree/initramfs-framework-ostree/files/ostree
@@ -1,8 +1,5 @@
 #!/bin/sh
 
-DISTRO='refkit'
-SCRIPT='/usr/bin/refkit-ostree'
-
 # We consider the ostree module always enabled if it is installed.
 ostree_enabled () {
     if [ -d /rootfs/ostree/deploy ]; then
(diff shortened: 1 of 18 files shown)


pohly:
@klihub: I squashed most of the commits into one, with you as original author mentioned in the commit and in the git commit itself, while the signoff is from me. Is that okay with you? We could also reset the author so that I get all the "git blame" for the files ;-}

New changes since you last looked at the commits:
- documentation
- same variable names (where applicable) as in meta-updater

pohly:
test this please

pohly:
Build https://ostroproject.org/jenkins/job/intel-iot-refkit_pull-requests/1722/ was okay, only the rebuild seems to have run into problems with testing on MinnowBoard.

retest this please

pohly:
@klihub I just noticed that you hadn't used your @intel.com email address for the commit. I'll change that to your work email address.

klihub:
Ah ok, thanks, and sorry about that. Part of the commits were done on my home machine and obviously I forgot to change my git-default mail address for the cloned repo.

pohly:
@klihub as part of the rebase I accidentally pushed two new commits that I was working on yesterday. They are okay to be merged, they just haven't been reviewed yet. Can you have a look at c96fece and ac2c72a and?

meta-refkit-core/recipes-ostree/initramfs-framework-ostree/files/ostree
```diff
@@ -29,27 +26,145 @@ ostree_debug () {
...
+    mount --move . /rootfs || return 1
+
+    cd - > /dev/null
+
+    return 0
+}
+
+# Prepare the initramfs environment for pivot-rooting into an ostree deployment.
+ostree_initramfs_prepare_root () {
+    local _entry_CURR _entry_PREV _path_CURR _path_PREV
```
pohly: @klihub it seems to me that this line doesn't actually do anything. These variables are set globally via the eval expressions in ostree_loader_entries(), so defining them as local in ostree_initramfs_prepare_root is redundant. It seems to work because "local" only declares them as local when defined, without actually defining the variables and thus shadowing the global ones - but I'm a bit fuzzy on the exact semantic, so my guess might be wrong.

Either way, this should get removed in a future cleanup commit. Agreed?

klihub: I agree that this should get removed in a cleanup commit. I almost have a reworked version of this which does away, among other things, with parsing by the shell script.

That said, the shell evaluation/scoping model is not quite that primitive. It's a variant of nested scoping where the closer nested scope shadows any further one. The shell code in there deliberately declares the variables local to limit their scope of visibility to itself and the functions it calls, ostree_loader_entries included, so effectively the eval there is setting a local variable upper in the call stack.

pohly: Thanks, learned something new today.

However, "local' isn't part of POSIX. ash from Busybox supports it, but who knows how complete the scoping support really is.

PR #214, opened 2017-06-28
flatpak-enable the stock gateway image

klihub:
This patch set enables flatpak support for the stock gateway image. IOW it essentially turns the stock refkit-image-gateway into refkit-image-gateway-flatpak-runtime, as long as the required flatpak distro features are enabled. Consequently, the patches also remove building and testing the explicitly spelled out flatpak variant of the same image in CI.


klihub:
test this please

PR #215, opened 2017-06-29
Combined PR

mythi:


mythi:
Merging despite one networking glitch in daft.

PR #195, opened 2017-06-16
esdk: build and test eSDK installers

mythi:

PR #217, opened 2017-06-30
doc: Image configuration methods

ipuustin:
Detail some methods for Refkit image configuration.

### DIFF
--- a/doc/howtos/image-configuration.rst
+++ b/doc/howtos/image-configuration.rst
@@ -1,0 +1,92 @@
+Configuring IoT Refkit images
+#############################
+
+Many packages in IoT Refkit come with a default set of configuration
+files. This document gives guidance for configuring a Refkit image with
+custom package configuration files.
+
+Using .bbappend files to override package configuration
+=======================================================
+
+If you have only a limited set of packages which you need to configure,
+you can create ``.bbappend`` files which append the recipes for each
+package which you want to modify. A usual way to change the
+configuration is to change the install phase to replace or remove the
+default configuration files.
+
+You can add the ``.bbappend`` files to your own customization layer and
+use ``do_install_append()`` for modifying package configuration.
+
+.. code::
+
+    do_install_append () {
+        install -d ${D}${sysconfdir}
+        echo "some configuration" > ${D}${sysconfdir}/config_file
+        ...
+    }
+
+Using virtual packages
+======================
+
+Some recipes in IoT Refkit and other Yocto layers have the configuration
+split into separate configuration packages. For example, see
+`nftables-settings-default recipe
+<../../meta-refkit-core/recipes-security/nftables-settings-default/nftables-settings-default_0.1.bb>`_.
+In these cases, if you want to replace the configuration fully, you can
+create your own configuration package (say ``nftables-settings-custom``)
+and just use it in place of the default configuration package.
+
+Other recipes, where having a configuration package is mandatory, take
+this further and use ``VIRTUAL_RUNTIME`` convention to enforce that a
+configuration package will be installed to the target. For example,
+consider `groupcheck recipe
+<../../meta-refkit-core/recipes-security/groupcheck/groupcheck_git.bb>`_,
+which has the following lines:
+
+.. code::
+
+    VIRTUAL-RUNTIME_groupcheck-settings ?= "groupcheck-settings-default"
+    RDEPENDS_${PN} += "${VIRTUAL-RUNTIME_groupcheck-settings}"
+
+If a configuration package (say ``groupcheck-settings-custom``) is made,
+it needs to be set to be the required configuration package. This is
+done by changing ``local.conf`` or other distro configuration file to
+override the particular ``VIRTUAL_RUNTIME`` variable:
+
+.. code::
+
+    VIRTUAL-RUNTIME_groupcheck-settings = "groupcheck-settings-custom"
+
+Image configuration during root filesystem creation
+===================================================
+
+An alternative to configuring individual packages is the configuration
+of the entire image during rootfs creation. The image recipes can be
+appended as any other recipe. This is the correct approach if you have
+several files in different packges which you need to change or if you
+just want to keep the configuration changes in one place.
+
+One practical way for implementing a certain configuration would be to
+add it as a ``.bbclass`` file, which would then be inherited by all
+images which require that configuration. For example, you can create
+``custom-config.bbclass`` file and add a Python task function to do the
+necessary configuration changes there:
+
+.. code::
+
+    python change_image_configuration () {
+        etcdir = oe.path.join(d.getVar("IMAGE_ROOTFS"), d.getVar("sysconfdir"))
+        ...
+    }
+
+    ROOTFS_POSTPROCESS_COMMAND_prepend = "change_image_configuration;"
+
+You can then inherit the class directly into your own image recipes.
+
+A drawback of this method is that the image is changed after the
+packages are installed. This means that you can't use package-based
+methods for updating devices, but should instead use `file- or
+block-based update mechanisms
+<https://wiki.yoctoproject.org/wiki/System_Update>`_. Supported system
+update mechanisms IoT Refkit are documented in `system-update.rst
+<../system-update.rst>`_.


doc/howtos/image-configuration.rst
```diff
@@ -0,0 +1,100 @@
+Configuring Refkit images
```
mythi: What's the intended audience for this doc? The guidance below does not say much about *where* to maintain the bbappends/bbs. Depending on the audience the answer will vary.

ipuustin: That's right. The intended audience is anyone who needs to build a customized Refkit image.

I think there would be a need for some general purpose document, which would be called "How to create a custom IoT distribution based on Refkit". That should include things like:
1. How to create an own layer for Refkit customization? Is it better to do a layer inside Refkit or add Refkit inside some other layer?
2. How to rebase when refkit changes? How to solve typical errors in the process?
3. How to properly do builds and push resulting updates to target devices?
4. How to develop and maintain your own recipes on top of Refkit?
And finally:
5. How to configure packages in Refkit? (this document)

I'm not sure if extending this PR would be the right place to do that. Instead We could do a set of documents answering these questions and link them together on top level.

doc/howtos/image-configuration.rst
```diff
@@ -0,0 +1,100 @@
...
+#########################
+
+Many packages in Refkit come with a default set of configuration files.
```
pohly: I think elsewhere we have used "IoT Refkit" as abbreviation.

doc/howtos/image-configuration.rst
```diff
@@ -0,0 +1,100 @@
...
+Using .bbappend files to override package configuration
+=======================================================
+
+If you have only a limited set of packages which you need to configure,
+you can create ``.bbappend`` files which append the recipes for each
+package which you want to modify. A usual way to change the
+configuration is to change the install phase to replace or remove the
+default configuration files.
+
+As explained in `yocto-compatible.rst <yocto-compatible.rst>`_, the
```
pohly: I must second mythi's comment here: this "Yocto Compatible" requirement only applies to .bbappends contained in refkit itself and thus to Refkit core developers. Users of refkit can do whatever they want in their custom distro (configuration) layer.

doc/howtos/image-configuration.rst
```diff
@@ -0,0 +1,100 @@
...
+necessary configuration changes there:
+
+.. code::
+
+    python change_image_configuration () {
+        etcdir = oe.path.join(d.getVar("IMAGE_ROOTFS"), d.getVar("sysconfdir"))
+        ...
+    }
+
+    IMAGE_PREPROCESS_COMMAND_prepend = "change_image_configuration;"
```
pohly: IMAGE_PREPROCESS_COMMAND runs very late, in do_image. Typically configuration changes (for example, the debug tweaks) are done earlier in do_rootfs via ROOTFS_POSTPROCESS_COMMAND. Please recommend to use that.

doc/howtos/image-configuration.rst
mythi: IMO, the general purpose document would be the 'development story' part. @juimonen has started the work already. I'm the owner of YOCTO 11495

We can keep this doc until other pieces are ready and then see if it makes sense to merge the docs or go with the linking approach.

doc/howtos/image-configuration.rst
```diff
@@ -0,0 +1,100 @@
...
+
+You can either inherit the file directly into your own image file or
+alternatively add to your distro configuration file
+
+.. code::
+
+    INHERIT_refkit-image-<profilename> += "custom-config.bbclass"
```
pohly: Are you sure that this works?

I suspect you meant:

    INHERIT_append_pn-refkit-image-<profilename> = " custom-config"

Note that this (currently) has the same problem with the virtual refkit-image-<profilename>-flatpak as with the virtual swupd images: because the PN is different in the virtual images, the override is only applied to the base image.

I had (finally, I should probably say) and idea how to address that: when creating the virtual image, we can simply extend the OVERRIDES to include not just PN, but also the base PN. However, meta-flatpak uses image-variant.bbclass, which currently cannot know whether that OVERRIDES change is desired, so the actual implementation will be a bit more difficult.

doc/howtos/image-configuration.rst
juimonen: We agreed with Ismo that I can somehow merge this to development story at some point

doc/howtos/image-configuration.rst
```diff
@@ -0,0 +1,100 @@
...
+
+to add the configuration to a certain image (replace ``<profilename>``
+with the name of the profile you want to configure).
+
+A drawback of this method is that the image is changed after the
+packages are installed. This means that you can't use package-based
+methods for updating devices (such as `dnf
+<https://fedoraproject.org/wiki/DNF?rd=Dnf>`_), but should instead use
+image-based update methods (such as `OSTree
+<https://ostree.readthedocs.io/en/latest/>`_).
```
pohly: Instead of mentioning only OSTree directly, perhaps better link to https://wiki.yoctoproject.org/wiki/System_Update and then for OSTree to the (now merged) doc/system-update.rst.

pohly:
On Fri, 2017-06-30 at 01:09 -0700, Ismo Puustinen wrote:
>      1. How to properly do builds and push resulting updates to target
>         devices?

It's worth noting that with OSTree, one can install the "official" build
and the locally on the device add a remote for a "custom" build
repository on the local LAN and switch back and forth between OS builds
from both repos.

Details to be determined...

doc/howtos/image-configuration.rst
ipuustin: Hmm, let's remove the whole "external inherit" then. I suspect anyone requiring image configuration functionality will anyway have their own image recipes.

doc/howtos/image-configuration.rst
ipuustin: Ok, makes sense.

doc/howtos/image-configuration.rst
pohly: Yes, custom image recipe is what we should promote.

BTW, my own proposal won't work either: I had forgotten that INHERIT is evaluated as part of parsing the base configuration, so it cannot depend on per-recipe overrides.

doc/howtos/image-configuration.rst
ipuustin: Ah. I was under the impression that all Yocto layers (including the custom distro configuration layer) are expected to be "Yocto Compatible" soon, and so including them would be expected not to change Refkit configuration as such. Does this depend on whether the customization layer is added inside refkit or if refkit is added inside the customization layer (meta-oe style)?

doc/howtos/image-configuration.rst
ipuustin: I'll change this to be consistent.

doc/howtos/image-configuration.rst
ipuustin: Ok, I'll change that.

mythi:
[skip-ci]

doc/howtos/image-configuration.rst
pohly: "Yocto Compatible" is a certificate that layers can apply for, but don't have to. It is meant as a mark of quality for layers that are meant to be used by others. It makes no sense for private or product-specific layers.

mythi:
All the review feedback is now addressed in the latest version. It seems generic enough for both refkit-developer and refkit-user to use. I'm OK with it at least. In the long run this content would be added in the 'development story' doc.

@pohly are you good (the review status is 'requesting changes')?

doc/howtos/image-configuration.rst
```diff
@@ -0,0 +1,92 @@
...
+
+    ROOTFS_POSTPROCESS_COMMAND_prepend = "change_image_configuration;"
+
+You can then inherit the class directly into your own image recipes.
+
+A drawback of this method is that the image is changed after the
+packages are installed. This means that you can't use package-based
+methods for updating devices, but should instead use `file- or
+block-based update mechanisms
+<https://wiki.yoctoproject.org/wiki/System_Update>`_. Supported system
```
pohly: Minor nitpick (not worth fixing?): a package-based system update is a file-based update mechanism. The key point of those mechanisms is that they operate on the entire rootfs. A better wording might be "use an atomic update mechanism which treats the entire modified rootfs as a single entity".

mythi: Thanks. I added your feedback in PR #195 commit 3a3f142.

mythi:
merged part of #195

PR #219, opened 2017-06-30
log command output

pohly:
With the recently enhanced runCmd/bitbake() in OE-core master we can get
output from the invoked command passed through the logger that each test
has, which provides much nicer output while a test runs: instead of of test
that just seems to hang, one gets a constant stream of messages with time
stamps.

The command also gets logged, so we can remove some extra output which now
is a bit redundant.

### DIFF
--- a/meta-refkit-core/lib/oeqa/selftest/cases/image_installer.py
+++ b/meta-refkit-core/lib/oeqa/selftest/cases/image_installer.py
@@ -59,9 +59,7 @@ class ImageInstaller(OESelftestTestCase):
             # temporarily in a way that it overrides some other IMAGE_MODE setting in local.conf.
             self.append_config('IMAGE_MODE_forcevariable = "development"')
             targets = 'refkit-installer-image ovmf swtpm-wrappers-native'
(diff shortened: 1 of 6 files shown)


okartau:
something in this change (but what?) contributed so that bitbake found zero usable elements from sstate
and rebuilt everything, taking 1h more in build stage than PR builds typically

pohly:
OE-core was updated. I'm not sure what changed, but probably something in that update.

pohly:
retest this please
test this please

okartau:
test this please

pohly:
@okartau any idea why you can start retesting and I can't?

okartau:
no explanation. The internal glitches of that plugin have remained mysteries so far, no debug/log.
I don't believe it's user-based, as I have also been unable to trigger from comment, and you have been able to trigger. I believe more in theory that 1st attempt sometimes fails. We just have seen few times that you make 1st try and then I see it's not triggering, make 2nd, at that works.

mythi:
@pohly could you rebase this

PR #222, opened 2017-07-02
meta-refkit-core: switch to C-based refkit-ostree for the initramfs ostree module, and ostree updater.

klihub:
This patch set switches refkit-ostree, the initial ostree wrapper/helper and updater script, to a C-based implementation. The refkit ostree initramfs module has been updated to use this for parsing the boot loader configuration files and pick the latest ostree deployment to boot into.

The patch set also enables automatic ostree-based image updates, updating the UEFI combo app if necessary, and activating the changes via a reboot. On the first boot after an update, a series of checks are performed to decide if the update is considered successful and working. Additional checks can be easily hooked to the existing set (for instance by services or applications, if necessary). The current set of checks do nothing, just marking the update successful.

Note that if the updater service (refkit-update.service) is enabled (which it is by default), entities that need to prevent (= delay) the shutdown to a more convenient time in the future should use systemd-inhibit, or the corresponding systemd(-logind) interfaces, for doing so. For instance, if you have an interactive shell (or a login session via ssh) while the updater is running and you don't want the system to be rebooted under you in case an update gets pulled in, you should do a

```
    systemd-inhibit --what=shutdown $SHELL
```

Once you're done, you can simply exit the shell to let any potential reboot/updates to proceed.

To start using/testing ostree-based updates, you need to
1) enable ostree
2) expose the generated ostree archive-z2 repository over HTTP/HTTPS
3) configure the ostree remote to point to your exposed repository

I usually do this in the following way, although there are plenty of other ways, too (for instance, directly setting the ostree remote to your development/build machines IP address or host name in which case you don't need to fiddle with /etc/hosts on your test box).

- enable ostree support in local.conf (and enable password-less root logins):
```
REFKIT_IMAGE_EXTRA_FEATURES += "ostree debug-tweaks"
```
- override the default remote in local.conf, to use HTTP instead of HTTPS:
```
OSTREE_REMOTE = "http://update.example.org/ostree/"
```

- expose the ostree repo generated during the build using apache by adding this:
```
Alias "/ostree/" "<path-to-your-cloned-refkit-repo>/build/tmp-glibc/deploy/ostree-repo/"

<Directory <path-to-your-cloned-refkit-repo>/build/tmp-glibc/deploy/ostree-repo>
    Options Indexes FollowSymLinks
    Require all granted
</Directory>
```

- and finally on my test box running a now ostree-enabled image, I alias the remote to my host in /etc/hosts:
```
...
<build-machine IP address>    update.example.org              update
...
```

Then I just make sure my firewall lets HTTP traffic in, generate new builds and let my test box automatically pull the new builds in and reboot. If I need to, I use the systemd-inhibit trick above to prevent the box from rebooting when an update has been pulled in.


pohly:
Where is the refkit-ostree-initramfs.bb that gets referenced by the "supported recipes" update? Should it be in this PR? I don't see it under "files changed".

klihub:
Hmm... I squashed it today (maybe a bit overeagerly) into the first commit swallowing the code from the github repo. This is still somewhat WiP, so I can split that change out into a separate commit of its own.

I also squashed the UEFI combo-app updating post-apply hook into the same commit. I think (as a functional change) that could/should also be kept as a separate commit of its own. What do you think ?

klihub:
Ah, I see what you mean. Thanks for spotting it. I'll fix that. Shouldn't have added the package to the recipe list. Weekend-commits...

klihub:
Yeah, I tend to agree... mostly about the fact that it should be easy to adjust (mostly by adding to) the actions we perform during the ostree rootfs preparation... which is an automatic side-effect if the core of the actions are written as a shell script. Our customers (and we ourselves as we're pretty much still prototyping) might need to alter the setup sequence.

Extensibility could be achieved in other trivial ways too, but using a shell script is probably the most straightforward. That's why we do have the shell-parsing options (--latest-entry --shell) in place, because that was my original plan. While I was at it I just got curious how it would look like to have the core of the actions implemented in C so I wrote it to find out.

Just give me a sec, and I roll a shell equivalent.

klihub:
There you go...

pohly:
Doing the parsing in C and the rest in shell looks like a good approach.

pohly:
Can the building of the C project be simplified? Do we really need shave and/or a full copy of it, for example?

klihub:
Sure. I axed out support for/usage of shave.

meta-refkit-core/recipes-images/images/refkit-initramfs.bb
```diff
@@ -55,7 +55,7 @@ FEATURE_PACKAGES_debug = "initramfs-module-debug"
 IMAGE_FEATURES += " \
     ${@bb.utils.contains('DISTRO_FEATURES', 'ostree', 'ostree', '', d)} \
 "
-FEATURE_PACKAGES_ostree = "initramfs-framework-ostree"
+FEATURE_PACKAGES_ostree = "initramfs-framework-ostree refkit-ostree-initramfs"
```
mythi: scratch that. I found it.

meta-refkit-core/recipes-ostree/refkit-ostree/files/refkit-ostree/scripts/refkit-ostree
```diff
@@ -0,0 +1,714 @@
...
+    if ! root_uuid _uuid; then
+        fatal_msg "failed to determine root partition UUID"
+    fi
+
+    _root=${_path_CURR}
+    _arch=$(uname -m)
+
+    case $_arch in
+        *64) _active=bootx64.efi;;
+        *)   _active=bootx86.efi;;
```
mythi: minor: non-64bit becomes ```bootia32.efi```

mythi:
overall, +1 from me.

meta-refkit-core/recipes-ostree/refkit-ostree/files/refkit-ostree/scripts/refkit-ostree
klihub: Fixed it.

klihub:
Updated to use pre-generated keys from the repository.

klihub:
test this please

klihub:
@okartau Olev, could you run on the CI server the following command and provide the output: gpg --homedir /srv/jenkins/workspace/ci-2017-07-07_08-30-22-build-2206/build/tmp-glibc/deploy/gnupg --list-keys ?

okartau:
Note that this is not exactly from builder context (which is inside docker container) but as that container is gone, I tried to run it in builder host context, there happens to be gpg as well.
I hope that does not make big difference.

jenkins@r630-6:~/workspace/ci-2017-07-07_08-30-22-build-2206> gpg --homedir /srv/jenkins/workspace/ci-2017-07-07_08-30-22-build-2206/build/tmp-glibc/deploy/gnupg --list-keys 
/srv/jenkins/workspace/ci-2017-07-07_08-30-22-build-2206/build/tmp-glibc/deploy/gnupg/pubring.gpg
-------------------------------------------------------------------------------------------------
pub   2048D/9A35FC2F 2017-06-30
uid       [ultimate] Signing Key <refkit-development-signing@key>
sub   2048g/785DFBA1 2017-06-30

pub   2048D/49785312 2017-07-07
uid       [ultimate] Signing Key <refkit-development-signing@key>
sub   2048g/777D0F58 2017-07-07

okartau:
build 2206 really had ostree segfaulting as was mentioned in job log.
Here is piece from builder syslog:

Jul 07 19:01:51 r630-6 kernel: ostree[95570]: segfault at c557b018 ip 00007f1842d3ec09 sp 00007ffcc4812ab0 error 4 in libc-2.25.so[7f1842cc8000+196000]
Jul 07 19:01:53 r630-6 systemd-coredump[95787]: Process 95570 (ostree) of user 600 dumped core.

okartau:
there is ostree coredump stored on builder as well:

r630-6:/var/lib/systemd/coredump # ls -lh
total 1.1M
-rw-r-----+ 1 root root 1.1M Jul  7 19:01 core.ostree.600.13956070a7514d88aa50c87ca8ebc015.95570.1499443311000000.xz

klihub:
Ok, thanks. That keyring dump is indicative enough as well.

pohly:
On Fri, 2017-07-07 at 12:54 -0700, Krisztian Litkey wrote:
> @okartau Olev, could you run on the CI server the following command
> and provide the output: gpg
> --homedir /srv/jenkins/workspace/ci-2017-07-07_08-30-22-build-2206/build/tmp-glibc/deploy/gnupg --list-keys ?

Do we end up calling gpg from the build host? We need to avoid that and
should build gpg-native instead.

klihub:
Hmm... I don't think that's the problem. We do build gnupg-1.x native so we end up using that one.

I think the problem is that somehow I broke the gpg serialization hack (or maybe it was already broken) and we incorrectly end up generating (from parallel tasks) two keys for the same key ID. While gpg survives that easily, ostree/gpgme crashes when/if that happens.

This should not happen if everything worked as intended (we should import the pre-generated key once and be done with it)... need to figure out where things go wrong.

klihub:
You can also see it from the key dates. One is pre-generated but we end up generating another one anew. I screwed up the protection logic against that...

pohly:
On Sat, 2017-07-08 at 05:59 -0700, Krisztian Litkey wrote:
> You can also see it from the key dates. One is pre-generated but we
> end up generating another one anew. I screwed up the protection logic
> against that...

I hadn't looked at the actual code change and had assumed that you had
removed refkit-signing-keys.bbclass, as discussed with Mikko.

Now I had a look and see that refkit-signing-keys.bbclass is still
there, with some additional logic to import pre-generated keys. Is that
really necessary?

Can't we commit the output of key generation with GPG 2.x into our repo
and then point gpg-native at that directory whenever we need to do
signing?

Something like 

OSTREE_GPGDIR ?= "${@ '' if (d.getVar('IMAGE_MODE') or 'production') == 'production' else '${META_REFKIT_CORE_BASE}/files/gpg-signing'}
OSTREE_GPGID ?= "${@ '' if (d.getVar('IMAGE_MODE') or 'production') == 'production' else 'refkit-development-signing@key' }"

in ostree-image.bbclass and something similar for flatpak.

klihub:
Well, in principle I think we could commit a full pre-populated GPG-directory to our repo and point ostree at it. At least from the top of my head, I cannot come up with anything why that should not work.

klihub:
I'll try it out in practice and let's see if there's somethink we didn't think of...

klihub:
Ok, now both ostree and flatpak have been changed to use a pre-generated keyring directly from the repository and the usage of the key generating/importing bbclass has been removed.

klihub:
Okay, I cleaned up the history, squashed all sets of logically related commits, and added a script and instructions in the appropriate commit message about how to generate and take into use new signing keys.

meta-refkit-core/recipes-ostree/refkit-ostree/files/refkit-ostree/README.md
```diff
@@ -0,0 +1,2 @@
+This package contains extra binaries and scripts used to help booting
+and keeping up-to-date devices running IoT Reference Kit OS images.
```
pohly: It's "IoT Reference OS Kit".

meta-refkit-core/recipes-ostree/refkit-ostree/files/refkit-ostree/src/refkit-ostree.c
```diff
@@ -0,0 +1,1642 @@
+/*
+ * Copyright (c) 2017, Intel Corporation
```
pohly: I have not reviewed the new C code in detail. Has anyone else reviewed it before and was static code analysis done for it?

meta-refkit-core/lib/oeqa/selftest/cases/refkit_ostree.py
```diff
@@ -99,7 +99,7 @@ def create_httpd():
 #exec socat 2>>/tmp/ostree.log -D -v -d -d -d -d STDIO TCP:localhost:%d
 ''' % port)
 
-            cmd = '''ostree config set 'remote "updates".url' http://%s && refkit-ostree update''' % self.OSTREE_SERVER
+            cmd = '''ostree config set 'remote "updates".url' http://%s && refkit-ostree-update -O -P "/usr/share/refkit-ostree/hooks/post-apply.d/00-update-uefi-app"''' % self.OSTREE_SERVER
```
pohly: Can you use the long version of the command line options (--one-shot and --post-apply-hook) here and explain in the source code (instead of the commit message) why it is done this way?

The comment about "only run UEFI app update hook in update selftest" will be easier to understand when it also mentions what is *not* run, i.e. the automatic reboot hook.

meta-refkit-core/classes/ostree-image.bbclass
```diff
@@ -66,23 +66,17 @@ OSTREE_COMMIT_SUBJECT ?= 'Build ${BUILD_ID} of ${PN} in ${DISTRO}'
...
-# Beware that generating keys can hang when the kernel runs out
-# of entropy for /dev/random, and the current refkit-signing-keys.bbclass
-# does not handle setting REFKIT_SIGNING_KEY in image recipes like
-# we do here. TODO: replace with pre-generated keys?
-OSTREE_GPGDIR ?= "${REFKIT_SIGNING_GPGDIR}"
+OSTREE_GPGDIR ?= "${@ '' if (d.getVar('IMAGE_MODE') or 'production') == 'production' else '${META_REFKIT_CORE_BASE}/files/gnupg' }"
 OSTREE_GPGID_DEFAULT = "${@d.getVar('DISTRO').replace(' ', '_') + '-development-signing@key'}"
-REFKIT_SIGNING_KEYS_append = " ${OSTREE_GPGID_DEFAULT}"
 OSTREE_GPGID ?= "${@ '' if (d.getVar('IMAGE_MODE') or 'production') == 'production' else '${OSTREE_GPGID_DEFAULT}' }"
-inherit refkit-signing-keys
```
pohly: doc/system-update.rst needs to be updated as part of this commit, because it still describes the old behavior under "OSTree Usage".

meta-refkit-core/recipes-ostree/refkit-ostree/refkit-ostree_0.0.bb
```diff
@@ -0,0 +1,42 @@
...
+S = "${WORKDIR}/${PN}"
+
+PACKAGES += "${PN}-initramfs"
+
+FILES_${PN}-initramfs = " \
+    ${bindir}/refkit-ostree \
+"
+
+FILES_${PN} = " \
+    ${bindir}/refkit-ostree-update \
```
klihub: I think it's better to do it now rather than later... that way we can make sure it gets properly done and not forgotten/omitted becuase of some other hurries. So let's do it now... I'll split out the most useful bits from that commit message and add it to doc/system-update.rst and update the PR once done.

meta-refkit-core/recipes-ostree/refkit-ostree/refkit-ostree_0.0.bb
```diff
@@ -0,0 +1,42 @@
+SUMMARY = "IoT RefKit ostree helper, scripts, services, et al."
```
klihub: I guess it is possible to switch back from a non-versioned to a versioned recipe in case it turns out later we want/need to start versioning, right. Then, we can safely change to a non-versioned recipe. I'll include that in an update PR.

meta-refkit-core/lib/oeqa/selftest/cases/refkit_ostree.py
klihub: Ok, I'll update.

meta-refkit-core/recipes-ostree/refkit-ostree/refkit-ostree_0.0.bb
klihub: I split out the useful bits to a separate commit, adding a new (initial version of an) OSTree howto and updating the existing system-update doc with slight adjustments and a pointer to the howto. Couldn't really squash the doc bits to the original C-ifying/repo-swallow commit as it would have then been a forward-reference to the situation only in place after the following initramfs-updating commit which is better to keep separate. This is now quite a reasonable compromise.

meta-refkit-core/recipes-ostree/refkit-ostree/refkit-ostree_0.0.bb
klihub: Switched to non-versioned.

meta-refkit-core/recipes-ostree/refkit-ostree/files/refkit-ostree/README.md
klihub: Thanks, fixed.

meta-refkit-core/classes/ostree-image.bbclass
klihub: Updated docs in the commit.

meta-refkit-core/recipes-ostree/refkit-ostree/files/refkit-ostree/src/refkit-ostree.c
klihub: The latest version of the code has not been run through yet. I'll try to get that done. Our standard go-to-guy for static scans is on vacation. I will need to rebootstrap my you-know-which-scanner setup from scratch in my development machine and I suspect that will take a considerable amount of time. It'd be nice to get this thing moving forward already before and address any potential issues from the static analysis in a separate commit and PR.

klihub: I have run static analysis locally on my machine on the latest code (with full defect detection enabled). It did not discover anything relevant.

doc/howtos/OSTree.rst
```diff
@@ -0,0 +1,127 @@
...
+
+    REFKIT_IMAGE_EXTRA_FEATURES += "ostree"
+
+To use your GPG signing key pair for signing and verifying the OSTree
+repository commits, assuming your keys are in the keyrings in <gpg-home>
+with key ID release@example.org, add the following to your local.conf or
+other suitable global Yocto configuration file::
+
+    OSTREE_GPGDIR = "<gpg-home>"
+    OSTREE_GPGID  = "release@example.org"
```
pohly: I think it's worthwhile to add a note here that this is not necessary when building refkit images in the default development mode and why, and to explain the implications (image update not secure).

doc/howtos/OSTree.rst
```diff
@@ -0,0 +1,127 @@
+Handling System Updates Using OSTree
+####################################
+
+IoT Reference OS Kit ('refkit') has support for updating devices running
+one of the refkit profile images using an HTTP/HTTPs server and OSTree.
+Ostree is a library and suite of command line tools, largely inspired by
```
klihub: I'll update that.

doc/howtos/OSTree.rst
```diff
@@ -0,0 +1,127 @@
...
+Now assuming, you did not change the location, and you use Apache for
+service HTTP/HTTPS requests, you can expose this repository with Apache
+by adding the following to your Apache configuration::
+
+    Alias "/ostree/" "<path-to-intel-iot-refkit>/build/tmp-glibc/deploy/ostree-repo/"
+    
+    <Directory <path-to-inte-iot-refkit>/build/tmp-glibc/deploy/ostree-repo>
+        Options Indexes FollowSymLinks
+        Require all granted
+    </Directory>
```
klihub: Definitely the ostree trivial-http needs to be mentioned and while at it I can mention the python one, too. I'll update the docs.

klihub:
Hmm, can these recent UNSTABLE test results be caused by a race between the refkit-(ostree-)update systemd service and the test case manually running refkit-ostree-update --one-shot, and the system service winning the race, doing a full update cycle, including the final reboot ? I think that is a possibility anyway, so IMO we need to stop the systemd service in the test case.

doc/howtos/OSTree.rst
klihub: Yes. I'd say when we switch from GPG signing to what Jussi have been cooking, dropping gpg1 and the rest would be a natural thing to do.

pohly:
On Tue, 2017-07-11 at 10:53 -0700, Krisztian Litkey wrote:
> Hmm, can these recent UNSTABLE test results be caused by a race
> between the refkit-(ostree-)update systemd service and the test case
> manually running refkit-ostree-update --one-shot, and the system
> service winning the race, doing a full update cycle, including the
> final reboot ? I think that is a possibility anyway, so IMO we need to
> stop the systemd service in the test case.

That sounds like a plausible explanation.

Can you include a fix in this PR?

pohly:
On Tue, 2017-07-11 at 17:38 +0000, Krisztian Litkey wrote:
> Definitely the ostree trivial-http needs to be mentioned and while at
> it I can mention the python one, too. I'll update the docs.

We just discussed on IRC (in some other context) that http.server isn't
particularly good at serving many requests. Better mention
http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/lib/oeqa/utils/httpserver.py

klihub:
If I can manage to wrap my head around the test case implementation well enough at this hour then I can try to roll a fix for it.

okartau:
UNSTABLE from post-build tests should be interpreted as FAILURE.
Why we see UNSTABLE now after recent merge is side effect from running post-build-tests and DAFT tests in parallel. There is override to UNSTABLE that was supposed to be for DAFT only, indicating issue in daft code itself, not tests. Now this override covers all tests.
It's not big issue, as for GH status, unstable also means failure, so mostly everything is ok,
But we need to think how to get rid of that side effect perhaps.

okartau:
off-topic for this PR, but continuing from my previous comment, should we make post-build test behave same way as DAFT is, so that error in tests (comes via xml files) will mark job FAIL, but error in running (comes as nonzero status from process) will mark job UNSTABLE (i.e. problem with testing code). If that is way to go, then current Jenkinsfile code is already good as it does just this.

klihub:
Well, anyway that UNSTABLE led us to find a potential real-world testing problem, which is hopefully plugged now (let's see if I managed to not screw it up). As for the UNSTABLE/FAIL vs. CI/DAFT... I don't know enough about CI and DAFT to form or state an educated opinion about this.

klihub:
test this please

meta-refkit-core/lib/oeqa/selftest/cases/refkit_ostree.py
```diff
@@ -70,6 +77,12 @@ def update_image(self, qemu):
         old_cwd = os.getcwd()
         server = None
         try:
+            # We need to stop the refkit-udpate systemd service before starting
+            # the HTTP server (and thus making any update available) to prevent
+            # the service from racing with us and potentially winning, doing a
+            # full update cycle including a final reboot.
+            self.stop_update_service(qemu)
```
pohly: We could also have created the image with the update service disabled in a ROOTFS_POSTPROCESS_COMMAND, but I like this solution better because it also covers the systemctl runtime operation.

doc/howtos/OSTree.rst
```diff
@@ -0,0 +1,153 @@
...
+ * expose the OSTree repository of your builds over HTTP/HTTPS
+ * point your clients to the exposed OSTree repository
+
+To enable OSTree support, turn on the 'ostree' image feature. You can
+do this by incuding the following configuration snippet in your local.conf
+or other suitable global Yocto configuration file::
+
+    REFKIT_IMAGE_EXTRA_FEATURES += "ostree"
+
+You can generate a pair of signing keys with the following command::
```
pohly: I still think we should mention here that generating keys is not needed for "development" images.

doc/howtos/OSTree.rst
```diff
@@ -0,0 +1,153 @@
...
+following commands::
+
+    cd build/tmp-glibc/deploy
+    ln -sf ostree-repo ostree
+    ostree trivial-httpd --port 80
+
+A third alternative is use a simple Python HTTP server, for instance the
+one from the Yocto project at::
+
+    http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/lib/oeqa/utils/httpserver.py
```
pohly: This URL gets rendered as verbatim text. Turning "the one from the Yocto Project" into a link would be more useful.

Also note the spelling: "Yocto Project", not "Yocto project".

doc/howtos/OSTree.rst
```diff
@@ -0,0 +1,153 @@
...
+
+which is also available in the refkit source tree as::
+
+    openembedded-core/meta/lib/oeqa/utils/httpserver.py
+
+Now with teh above configuration in place, and an HTTP server running,
```
pohly: s/teh/the/

doc/howtos/OSTree.rst
```diff
@@ -0,0 +1,153 @@
...
+need to
+
+ * enable OSTree support for your builds
+ * ideally generate and use a pair of signing keys for your builds/updates
+ * expose the OSTree repository of your builds over HTTP/HTTPS
+ * point your clients to the exposed OSTree repository
+
+To enable OSTree support, turn on the 'ostree' image feature. You can
+do this by incuding the following configuration snippet in your local.conf
+or other suitable global Yocto configuration file::
```
pohly: Remove "Yocto" here - "Yocto Project" is a brand that isn't applicable here.

doc/howtos/OSTree.rst
pohly: Actually, "httpserver.py", is part of OpenEmbedded core more than it can be said to be part of the Yocto Project.

Better just say "OpenEmbedded core" and link to http://git.openembedded.org/openembedded-core/tree/meta/lib/oeqa/utils/httpserver.py

doc/howtos/OSTree.rst
klihub: Removed Yocto from this and a few similar occurences.

doc/howtos/OSTree.rst
klihub: Added a note about using pregenerated keys for development images.

doc/howtos/OSTree.rst
klihub: Thanks for spotting. Fixed.

pohly: Please push the fixed version.

PR #223, opened 2017-07-02
meta-flatpak: fix bashism when incrementing 'slept' var.

okartau:


okartau:
hm, pyshlex.py does not like shell expression slept=$((slept+1)) that IMHO should be valid shell syntax working in both bash and dash

pohly:
On Sun, 2017-07-02 at 11:15 -0700, okartau wrote:
> hm, pyshlex.py does not like shell expression slept=$((slept+1)) that
> IMHO should be valid shell syntax working in both bash and dash

That's a pending enhancement:
https://bugzilla.yoctoproject.org/show_bug.cgi?id=11314

Use slept=`expr $slept + 1`

mythi:
Merging for now to get ubuntu builds ok

PR #224, opened 2017-07-03
refkit: reconfigure meta-ros in refkit-config.inc

pohly:
Setting ROS_USE_PYTHON3 in refkit-config.inc is better because:
- other distros may also need that setting
- all content re-configuration is one one place and thus
  easier to review


rojkov:
Looks good to me

PR #229, opened 2017-07-05
oe-selftest: Added gateway profile on refkit-license-check

YinThong:
Create license check for refkit gateway profile in oe-selftest.
In order to work with secure boot requirement. reduce the need for (L)GPLv3 components.
Add this checking on os-selftest script can easy detect the GPLv3 package.

[YOCTO #11494]

Signed-off-by: Choong Yin Thong <yin.thong.choong@intel.com>

### DIFF
--- a/meta-refkit-core/lib/oeqa/selftest/cases/refkit_license_check.py
+++ b/meta-refkit-core/lib/oeqa/selftest/cases/refkit_license_check.py
@@ -88,15 +88,13 @@ class LicensingTest(OESelftestTestCase):
 
         return os.path.join(manifestdir, "package.manifest")
 
-    def test_check_computervision_licensing(self):
+    def test_check_computervision_licensing(self, test_image='refkit-image-computervision'):
 
         """ Check that computer vision production build image can be
             made without using GPLv3 family licenses in any component.
         """
 
-        test_image = 'refkit-image-computervision'
-
-        print("test_check_computervision_licensing")
+        print("testing: %s" % test_image)
 
         # Create the test image (rootfs is enough).
         print('Building test image (%s)...' % test_image)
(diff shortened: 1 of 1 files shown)


YinThong:
This bug still cannot close because i only update the license check script. There is GPLv3 license detected on flatpak in refkit gateway profile. Continue check is there still got other impacted library.

meta-refkit-core/lib/oeqa/selftest/cases/refkit_license_check.py
```diff
@@ -88,15 +92,17 @@ def _get_latest_manifest(self, imagename, deploydir):
 
         return os.path.join(manifestdir, "package.manifest")
 
-    def test_check_computervision_licensing(self):
+    def test_check_computervision_licensing(self, test_image):
```
mythi: I'd suggest you add a default value for test_image here. With that, the printing below can be simplified to 

```python
print("testing: %s" % test_image
```

YinThong: Ok. The script result show there is bash package license issue in gateway. I will use computervision as reference because it look perfect.

pohly:
@YinThong shall this PR remain open as a test build until the known GPLv3 issues are addressed, or shall we close it for now? As it stands now, the test works as intended, but fails because of GPLv3 components.

You can always reopen it later, but beware that you need to do that *before* pushing something to your YinThong:bug_11494 branch. There's a quirk in GitHub that prevents re-opening for modified branches.

YinThong:
@pohly Sure just make it close for now.

pohly:
Let's reopen once the GPLv3 components are gone and the test succeeds.

PR #230, opened 2017-07-05
CI Jenkinsfile: Run post-build and DAFT tests in parallel

okartau:
CI Jenkinsfile: Run post-build and DAFT tests in parallel
We can't keep same docker container around for another node stmt,
but we can create a new docker container, which will
use same worktree from build stage. That allows to
put post-build tests into same set with daft-based tests
and run all of those in parallel, thus shortening
the job time by daft-tests duration (which is shorter than
post-build tests duration).

Second change in this PR is:
Stop calling workspace cleaning a stage.
    
After async. cleanup was introduced, deleteDir at start of
each job does practically nothing, but we keep it there just in case.
There is no point however put this short duration up on stages chart.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,9 +49,7 @@ try {
             ws("workspace/${slot_name}${ci_build_id}") {
                 builder_node = "${env.NODE_NAME}"
                 set_gh_status_pending(is_pr, 'Prepare for build')
-                stage('Cleanup workspace') {
-                    deleteDir()
-                }
+                deleteDir() // although dir should be brand new, empty just in case
                 stage('Checkout content') {
                     checkout_content(is_pr)
                 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -47,6 +47,7 @@ try {
     timestamps {
         node('rk-docker') {
             ws("workspace/${slot_name}${ci_build_id}") {
+                builder_node = "${env.NODE_NAME}"
                 set_gh_status_pending(is_pr, 'Prepare for build')
                 stage('Cleanup workspace') {
                     deleteDir()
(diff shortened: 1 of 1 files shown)


Arthil:
The GPIO pins were accidentally attached wrong way, when fixing a BBB problem (had to remove all the wires). It's fixed now. Test this please.

pohly:
@okartau do you still consider this a test build?

okartau:
This works well under regular conditions, however with higher load (i.e. queued jobs) this scheme
means that some next job's build stage will get it's turn on same builder before post-build testing starts,
causing this job post-build step to be queued on same builder.
This results in delayed run by next-job build stage.
This possibility has kept me from removing test-build label.
But OTOH, I dont have good plan to avoid such queue event from happening.
I don't want to start implementing some ugly-looking locking scheme.
Also, I dont want to add 2nd slot on all builders (which could run post-build tests in parallel).

Perhaps, we should just admit there is such a possibility that post-build test could run with delay,
and consider this as "hopefully rarely".
If mostly system is not under heavy load (no queue), 
then most of the times we could still see time savings from parallel run of daft and post-build.

pohly:
@olev let me rephrase, to ensure that I understand. So with the old scheme, the build and selftest phase were guaranteed to run on a build alone, with no other build running in parallel. Now, with the new scheme, there is a small chance that a second build starts on the same build host in parallel to the selftests of the first build?

I think that's acceptable. We can always reconsider when we notice issues.

okartau:
not in parallel, but it starts and delays post-build-test, which will start after it.
That happens because there is one slot on builder, it can run one job at a time.
I dont see big problem in that either, as original workspace will be intact, and delay
should not affect results in any way.
The delayed run is something not so nice here.
Or, some side effect or unwanted results from interleaved jobs that we cant foresee yet,
as we haven't tried those yet

pohly:
@okartau got it. Not running parallel is better than I thought, and getting delayed results for a PR is simply something we have to live with. It has not been particularly obvious (to the non-Jenkins experts among us, at least) how long it will take at any given moment for a new PR to be tested completely.

okartau:
ok, I also think same, so I will remove test-build label.
After all, in regular workflow we should almost never see the queued runs situation,
if we can leave the fake trigger-all issues behind.
I can grow safety marginal by adding few more workers, after we move to new CI master.

PR #232, opened 2017-07-07
CI: skip storing of plain .wic files

okartau:
publishing of plain .wic files to storage server
takes majority of transfer time and also space on server,
but there is no use for stored plain .wic files as wic.xz
are used.


okartau:
test it please

PR #233, opened 2017-07-07
stateless support

pohly:
This enables the "stateless" distro feature and the "stateless" image for
all refkit-image.bbclass images.

However, only the changes that do not require upstream source code patching
get enabled. For example, systemd configuration gets moved from /etc
entirely into /usr.

Advanced changes like allowing local user management separately from the
system users are not enabled because they depend on patches. Enabling those
changes would increase the risk that building IoT Refkit breaks when
OE-core gets updated, and at this point it is not certain whether that is a
risk worth taking.

The goals for merging these changes now are:
- making the technology available for experiments,
  without constantly having to update a separate branch
- simplifying the content of /etc for OSTree - it's not
  absolutely required as OSTree can handle /etc, but a bit nicer


okartau:
test it please

PR #235, opened 2017-07-10
CI: retry daft after unstable, fix bashism in summary, fix xml-reports handling, publish more sstate

okartau:
It makes sense to retry daft after unstable
result which means problem on tester.

2nd commit here is adding forgotten shebang in tester-create-summary.sh
which created trouble on Ubuntu where this script goes through dash

3rd commit fixes xml-reports handling which got partially broken with
recent parallel-testing change

4th commit is "publish sstate after post-build tests as well"
which was combined into this PR from PR#239

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -86,8 +86,10 @@ try {
                         throw e
                     } finally {
                         set_gh_status_pending(is_pr, 'Store images')
-                        params = ["${script_env}", "docker/publish-project.sh"].join("\n")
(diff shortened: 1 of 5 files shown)


okartau:
test it please

okartau:
This is not as simple that. we just call daft again  First testing round creates test results, 2nd adds another set, so double run will look like twice of cases. We need to delete results of 1st run if we want 2nd run look like single run

PR #240, opened 2017-07-11
TEST: build on Ubuntu

pohly:
Same as PR #220, just with less work to be done and extra debug output.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,7 +4,7 @@
 	branch = master
 [submodule "bitbake"]
 	path = bitbake
-	url = http://git.openembedded.org/bitbake
+	url = http://github.com/pohly/bitbake.git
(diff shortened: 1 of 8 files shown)


pohly:
Test this please.

pohly:
test this please

pohly:
Rebased and using a bitbake where the "bad file descriptor" error has special error handling.

okartau:
the 2nd commit of those you picked from PR#220 that adds packages, was updated in PR#220, this is needed to get xml reporting to work (the thing where pip needs to be used)

pohly:
We understand the root cause of the problem on Ubuntu (no_proxy=.<some domain> had no effect, see https://bugs.python.org/issue29142), so this PR can be closed.

PR #244, opened 2017-07-13
CI Jenkinsfile: streamline and shorten

okartau:
Combine two locks into one:
Move daft-tester xunit publish into finally-block,
net result should remain same.
That allows to combine two-in-row locked regions
and stop using 2nd lock.

2nd commit removes Github status setting code
which became unneeded/unused after move to different ghprb plugin


okartau:
test this please

okartau:
test this please

PR #251, opened 2017-07-20
meta-iotqa: modify wifi test to retry incase of error

Arthil:
DAFT test for testing wifi connection sometimes randomly fails. This fix forces the wifi test to retry at least 2 times before failing.

Signed-off-by: Christian da Costa <christian.da.costa@intel.com>

### DIFF
--- a/meta-iotqa/lib/oeqa/runtime/connectivity/wifi/wifi.py
+++ b/meta-iotqa/lib/oeqa/runtime/connectivity/wifi/wifi.py
@@ -134,14 +134,23 @@ class WiFiFunction(object):
             time.sleep(3)
         self.target_collect_info("route")
         assert status == 0, "Error messages: %s" % self.log
-
     def execute_connection(self, ap_type, ssid, pwd):
         '''
-        Do a full round of wifi connection without disable
+        Do a full round of wifi connection without disable (try 3 times)
         '''
-        self.enable_wifi()
-        self.connect_wifi(ap_type, ssid, pwd)
-        self.wifi_ip_check()
+        for i in range(3):
+            try:
+                self.enable_wifi()
+                self.connect_wifi(ap_type, ssid, pwd)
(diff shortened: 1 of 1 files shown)


okartau:
ok to test

Arthil:
I have already manually tested this in CI by intentionally triggering the an error. However, because the real problem happens randomly (and not very often), it is hard to test it in a real scenario.

meta-iotqa/lib/oeqa/runtime/connectivity/wifi/wifi.py
```diff
@@ -134,14 +134,23 @@ def check_internet_connection(self, url):
...
-        self.connect_wifi(ap_type, ssid, pwd)
-        self.wifi_ip_check()
+        for i in range(3):
+            try:
+                self.enable_wifi()
+                self.connect_wifi(ap_type, ssid, pwd)
+                self.wifi_ip_check()
+                break
+            except AssertionError:
+                if(i < 2):
```
pohly: Nitpick: this is Python code, there's no need for round brackets.

Arthil: I personally find it more clear with brackets, which is why I used them.

pohly: It's just not how it's usually done in Python, which makes the code look weird to people familiar with Python or (worse?) makes it look like the author is a C programmer who doesn't really know Python.

When in Rome, do as the Romans do.

rojkov: Even in C it would look weird.  `if` is not a function, but key word followed by an expression hence people put a whitespace between the key word and the rest of the statement to avoid confusions. 

Though I saw opposite coding styles - a whitespace between function name and brackets.

PR #253, opened 2017-07-20
refkit: upstream component update

pohly:
Several upstream changes made it necessary to adapt refkit:
- x86-base.inc switched from "live" to "hddimg" as default IMAGE_FSTYPE.
  We still don't want any of the functionality (and dependencies)
  from image-live.bblcass, so now all of its types get removed.
- One patch for meta-oic is now upstream.
- wic dependencies can (and need to, in order to avoid the
  large set of default dependencies) be set via WKS_FILE_DEPENDS.
- The "iotweb" collection was renamed to "iot-web" to match the
  meta-iot-web layer name.


pohly:
@rojkov do this changes make sense? Okay to merge?

rojkov:
Yes, looks good to me

PR #255, opened 2017-07-21
meta-iotqa: add comments to opencv python interface test suite

epyeoh:
Add comments to the main test suite and each
test cases.

Signed-off-by: Yeoh Ee Peng <ee.peng.yeoh@intel.com>

### DIFF
--- a/meta-iotqa/lib/oeqa/runtime/multimedia/opencv/opencv_cv2.py
+++ b/meta-iotqa/lib/oeqa/runtime/multimedia/opencv/opencv_cv2.py
@@ -2,7 +2,18 @@ import os
 from oeqa.oetest import oeRuntimeTest
 
 class OpenCVTest(oeRuntimeTest):
-
+    '''
+    This test suite tests the 3 main OpenCV components 
+    (CV, HighGUI, CXCore). Tests for Machine Learning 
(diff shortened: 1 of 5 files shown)


epyeoh:
@pohly I had added the comment / description for opencv test case. Please let me know if this is ok.

okartau:
test this please
failure was wifi-related, likely random

meta-iotqa/lib/oeqa/runtime/multimedia/opencv/opencv_cv2.py
```diff
@@ -56,28 +64,29 @@ def setUp(self):
                 OpenCVTest.cv_image_src),
             OpenCVTest.cv_image_src_target)
 
+    # Test that OpenCV can read image file from file system
```
pohly: This could also be done as docstring, i.e.
```
def test_cv_read_image(self):
    """Test that OpenCV can read image file from file system"""
```

But that doesn't matter in practice, so don't bother changing it.

epyeoh: Thanks for sharing, will look into it.

PR #258, opened 2017-07-25
Update whitelist layer reference of xf86-video-mga

jlaako:
Due to recent BSP changes, whitelist layer reference of xf86-video-mga
package needs to be updated.


pohly:
Actually, after thinking about this dependency on meta-oe some more and looking at the reason why we depend on xf86-video-mga, I've come to the conclusion that this is probably a bug in meta-intel. I've filed https://bugzilla.yoctoproject.org/show_bug.cgi?id=11837

pohly:
Superseeded by PR #259?

PR #259, opened 2017-07-25
CI: build also x11 and wayland images

pohly:
Regressions have not been noticed unless someone manually built images with
graphics. We need to do at least build testing, and ideally produce images
that then can also be tested with DAFT.

multiconfig enhancements for supportedrecipes are needed to catch the usage of xf86-video-mga@openembedded-layer.

### DIFF
--- a/docker/post-build.sh
+++ b/docker/post-build.sh
@@ -48,7 +48,7 @@ bitbake -S none ${_images}
 # Check linux-intel specifically in addition to images, because it did
 # rebuild at some point and even though bitbake-diffsigs should
 # recurse to it, that's not guaranteed to work.
-for target in linux-intel ${_images}; do
+for target in intel-linux ${_images}; do
   if ! bitbake-diffsigs -t $target do_build; then
     echo "$target: nothing changed or bitbake-diffsigs failed"
   fi
--- a/docker/build-project.sh
+++ b/docker/build-project.sh
@@ -100,12 +100,45 @@ fi
 # to run. This makes it possible to investigate signature changes in post-build.sh.
 bitbake -S none ${_bitbake_targets}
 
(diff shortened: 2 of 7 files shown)


pohly:
The build included meta-oe/recipes-graphics/xorg-driver/xf86-video-mga_1.6.5.bb, which should have failed because the recipe isn't from the expected layer. It succeeded, so I need to check whether supported-recipes.bbclass works as it should for multiconfig.

pohly:
@jlaako Can you review this?

Building X11 and Wayland images added about 11 minutes to the overall build time when reusing existing sstate (https://intel-iot-refkit.01.org/job/intel-iot-refkit_pull-requests/103/consoleFull from 00:45:04.763 till 01:06:01.376).

Note that I chose a different solution for xf86-videa-mga: instead of letting bitbake pick the (less tested?) version from meta-oe, I'm excluding that version from the build.

pohly:
This is no longer a test build. I propose to merge it.

docker/post-build.sh
```diff
@@ -48,7 +48,7 @@ bitbake -S none ${_images}
 # Check linux-intel specifically in addition to images, because it did
 # rebuild at some point and even though bitbake-diffsigs should
 # recurse to it, that's not guaranteed to work.
-for target in linux-intel ${_images}; do
+for target in intel-linux ${_images}; do
```
jlaako: Kernel package is called linux-intel, we also support linux-intel-rt as well as linux-yocto and linux-yocto-rt, the same ones supported by meta-intel BSP:
http://git.yoctoproject.org/cgit/cgit.cgi/meta-intel/tree/common/recipes-kernel/linux

pohly: This is debug code. It doesn't need to work for all of our kernels, only the one that we commonly build.

PR #262, opened 2017-07-27
CI: remove sstate debug code

pohly:
We no longer have unexpected rebuilds, so we can remove the debug code
that was added for debugging those rebuilds. Makes the build a bit
faster again.


okartau:
looks good!

PR #263, opened 2017-07-28
[doc only] README: delete misspelled hyperlink tag

okartau:
[skip-ci]

This tag is misspelled so it does not work.
Linking to this chapter works nevertheless, because
of implicit links to chapter headings.

(If we want to have explicit link as was original intention,
we can put this line back in corrected form)


okartau:
this PR also verified skip-ci tag works as expected now (that we switched to another GH PR plugin in CI)

PR #264, opened 2017-07-28
CI : update exclude set, publish and test multiple configs

okartau:
commit 1: update exclude set of tarball creation
Changes in build and test setup have caused leaking some extras
in source tarball which don't belong there,
like build.pre, buildhistory-*,
making tarball size 27MB instead of 12MB.

commit 2: publish multiple config results

commit 3: CI tester: added deploy path to tester params

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -134,9 +134,9 @@ try {
     test_targets = testinfo_data.split("\n")
     for(int i = 0; i < test_targets.size() && test_targets[i] != ""; i++) {
         def one_target_testinfo = test_targets[i]
-        def test_device = one_target_testinfo.split(',')[4]
-        def test_machine = one_target_testinfo.split(',')[3]
-        def img = one_target_testinfo.split(",")[0]
+        def test_device = one_target_testinfo.split(',')[5]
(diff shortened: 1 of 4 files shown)


okartau:
as expected, in this PR build tarball size changes
from 28417653
to     12507340

and contents after unpacking look reasonable

pohly:
@okartau this is not a test build anymore, right?

okartau:
yes, in the sense that this combination of changes is a working solution. But one can discuss is this good enough/best solution. I don't like very much hard-coded  "glibc" on few paths. And cutting of "tmp-" from glibc paths, is this good solution. And the fact that testsuite is only present in glibc area, but is cross-used for other multi-config areas testing.

pohly:
Yes, review of the actual change is still needed. Usually I wait with that until the original author is done with the "test" phase of the PR, hence my question.

pohly:
Regarding publishing the test suite for the default configuration: that's something for QA to decide upon. It might or might not make sense. We can keep it as it is for now and if it needs to be changed, do that together with adding tests for the non-default images.

okartau:
this can be closed as contents were merged in PR#267

PR #267, opened 2017-07-31
CI: update exclude set, publish and test multiple configs, "headless" main config

pohly:
This extends PR #264.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -134,9 +134,9 @@ try {
     test_targets = testinfo_data.split("\n")
     for(int i = 0; i < test_targets.size() && test_targets[i] != ""; i++) {
         def one_target_testinfo = test_targets[i]
-        def test_device = one_target_testinfo.split(',')[4]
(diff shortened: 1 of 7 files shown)


okartau:
now this fails on tester images fetch, because I added path glibc/ in few places which latest commit
turned into headless-glibc. Either you need to change those places in this PR, or I pick 4th commit from here into my original PR and continue there

okartau:
do_testsdkext still thinks DEPLOY_DIR is tmp-glibc/deploy ?

pohly:
Hrrm, must be hard-coded somewhere.

pohly:
@okartau Now all code and test seems to be happy with tmp-headless/x11/wayland-glibc. Shall we merge it?

okartau:
yes, looks good. One of my changes left indents in not-so-clean state where I turned piece of code into a function in publisher. My plan was to keep the indents for better diff readability in that commit, then fix indents in separate commit. but thats cosmetic issue and can be changed later. Lets move forward and merge this.

PR #269, opened 2017-08-01
test build: meta-clang build failure

pohly:
The compiler-rt and libcxxes patches are now upstream in meta-clang, but updating meta-clang breaks beignet.


ipuustin:
Probably beignet is not ready for Clang 5 yet. In my opinion we shouldn't move to use Clang 5 either, since it doesn't appear to be officially released yet.

mythi:
Are there changes to get the compiler-rt and libcxxes patches backported in meta-clang/pyro?

mythi:
replaced by #291

PR #275, opened 2017-08-02
CI: expose daft error, stop storing licenses in PR case, support multiple mail recipient lists, misc.cleanup

okartau:
1st commit:  narrow archived xml file set to TEST-*.xml in Jenkinsfile
Mask used to be *.xml causing over-archiving of results.xml
from every tester with latest one remaining in archive,
which was misleading as it appears different each time because it 
typicaly comes from different tests.
results.xml is not used for anything so leave it out.

2nd commit: : Check for failure pattern in daft log file in tester-exec and show containing file.
Show DAFT-stage errors in main level job log instead of explanation remaining buried in 
one of many artifacts from DAFT,  which needs locating and opening one-by-one in job artifacts area.
Note that same error is typically shown in results-runtest.*.log
and in test_aft.*.log, we show results-runtest*.log which is shorter.

3rd commit: rsync can remove sent file via cmd line option, can drop explicit rm command

4th commit: stop publishing licenses/ subdir in PR job
as not used for anything, just waste  of bandwidth and storage

5th commit: we can shorten Jenkinsfile, empty string vars dont need declaring if they
get value by assignment later

6th comment adds one level of indirection in mail recipients, allows to maintain 
different mail recipient sets fordifferent job variants

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,9 +14,10 @@
 // more details.
 //
 
+// JOB_NAME expected to be in form <repo-name>_<variant>_<type> (master jobs)
+// or: in form <repo-name>_pull-requests (PR jobs)
 def is_pr = env.JOB_NAME.endsWith("_pull-requests")
 def target_machine = "intel-corei7-64"
-// JOB_NAME expected to be in form <layer>_<branch>
 def current_project = "${env.JOB_NAME}".tokenize("_")[0]
 def image_name = "${current_project}_build:${env.BUILD_TAG}"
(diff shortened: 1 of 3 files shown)


okartau:
test this please
nftables related fail in daft testing, not related to this PR

okartau:
test this please
nftables related fail in daft testing, not related to this PR

okartau:
test this please
mraa gpio related fail in daft testing, not related to this PR

okartau:
test this please, fetch failure in post-build

okartau:
test this please, keeps failing for unrelated reasons, now flatpak and usrmerge

PR #276, opened 2017-08-03
meta-iotqa: Add QA tests for OpenCL

epyeoh:
Add QA tests for OpenCL library and ICD files

### DIFF
--- a/meta-iotqa/lib/oeqa/runtime/multimedia/opencl/opencl.py
+++ b/meta-iotqa/lib/oeqa/runtime/multimedia/opencl/opencl.py
@@ -1,0 +1,7 @@
+from oeqa.oetest import oeRuntimeTest
+
+class OpenCLTest(oeRuntimeTest):
+    def test_opencl_app_can_execute(self):
+        (status, output) = self.target.run('/usr/bin/opencl-bench-opencl')
+        self.assertEqual(status, 0, msg="OpenCL error messages: %s" % output)
+
--- a/meta-iotqa/conf/test/refkit-image-computervision.manifest
+++ b/meta-iotqa/conf/test/refkit-image-computervision.manifest
@@ -1,6 +1,7 @@
 # Tests for computervision profile
 oeqa.runtime.multimedia.realsense.realsense_headless
 oeqa.runtime.multimedia.opencl.opencl_viennacl_1
+oeqa.runtime.multimedia.opencl.opencl
 oeqa.runtime.multimedia.opencv.opencv_dnn_1
 oeqa.runtime.multimedia.opencv.opencv_cv2
 oeqa.runtime.multimedia.gstreamer.gstreamer_cli


epyeoh:
@ipuustin @pohly these QA tests are ready for merging. Can you review this? Thank you!

ipuustin:
@Arthil @samniem: Could you verify that the test format is right and the tests really do get run?

meta-iotqa/lib/oeqa/runtime/multimedia/opencl/opencl.py
```diff
@@ -0,0 +1,10 @@
+from oeqa.oetest import oeRuntimeTest
+
+class OpenCLTest(oeRuntimeTest):
+    def test_opencl_has_library_file(self):
+        (status, output) = self.target.run('ls /usr/lib/libOpenCL*')
```
ipuustin: Maybe `test -f filename` would be a cleaner way to check for the existence of a file? `ls` promises to return non-zero if an error occurred, but I think it might not be defined in the spec that a missing file counts as an error when using wildcards.

epyeoh: Noted, will test the suggested way and update the code. Thank you!

samniem:
Looks good to me as well

meta-iotqa/lib/oeqa/runtime/multimedia/opencl/opencl.py
mythi: @epyeoh I don't see the feedback from @ipuustin is addressed. Do you plan to update the PR accordingly?

epyeoh:
@ipuustin @mythi  The objective of the test cases for OpenCL here is to first verify the needed library files are present inside the OS image before executing OpenCL executable. From reading information related to OpenCL, it stated that libOpenCL.so, also called OpenCL ICD Loader, does not contains any OpenCL implementation itself, but merely act as a dispatcher to real OpenCL implementations provided as OpenCL Installable Client Driver (ICD). Thus, the test case will first make sure the libOpenCL.so file exist and ICD files are inside the default directory (/etc/OpenCL/vendors/). 

In term of implementing the file check, I compared (a) test -f vs (b) ls or find. (a) test -f does not support regular expression and mainly for specific file name. Whereas (b) ls or find, support regular expression, eg. libOpenCL*. 
 
To verify that inside /etc/OpenCL/vendors contains .ICD files (intel-beignet-550x.icd, intel-beignet-570x.icd, intel-minnowmax.icd, intel-xxxx.icd). In this case, to use file -f require one to explicitly listing all .ICD file names. Instead, ls or find with regular expression or wildcard can be used to make sure there are .ICD files for the ICD Loader. 

Question (1): Do you have any inputs or concerns related to the objective: To verify that inside /etc/OpenCL/vendors contains .ICD files. If you agree with the objective, do you think using "ls" will meet this objective?

To verify that the libOpenCL.so file exist. Ideally using test -f <filename> is the best method. Problem is the major and minor version number behind the file name, eg. libOpenCL.so.1.0.0. To use "test -f" will require specific the major and minor version number and thus making the test case hard to maintain. 

Question (2): In this case, do you have any inputs or concerns on using ls or find instead of "test -f" to check for libOpenCL.so?

mythi:
@epyeoh I have one concern: if those files are missing, the existing viennacl or opencl-bench-opencl would fail. What's the idea of checking those files separately if we can see them missing in a functional test?

epyeoh:
Yes if any of the important files (.ICD or libOpenCL.so) are missing, the opencl executable will fail. When this test case failed, it leave little clue to the viewer and the viewer may need to spend more time to find the real root cause. 
Those lower level test cases that check each important files are designed to provide clear & specific message to let the viewer or maintainer know exactly which file is missing. Example if test_opencl_has_icd_file failed, the viewer know that .ICD file are missing and thus able to find root cause in shorter time.

mythi:
My preference is it's sufficient to print out the error output from the test binary. The check library part is already covered by running the test binary (the error indicates 'unable to load lib'). The .icd checking part is also not very portable/generic because it checks all refkit specific loaders.

epyeoh:
@mythi 
1) check library - noted, we shall remove the test if it is already enough test case to cover. 
2) check .ICD - yes it merely check all refkit specific loaders, we shall remove the test. 
Thank you for your inputs.  Let me change it and check-in the code.

epyeoh:
test this please
mraa gpio related fail in daft testing, not related to this PR

epyeoh:
test this please
it failed again on mraa gpio related testing, not related to this PR

epyeoh:
test this please
it failed again on mraa gpio related testing, not related to this PR
@okartau any idea why it is consistently failing on mraa gpio?

okartau:
3 days ago (on last week Fri) there was incident where PR#254 testing
caused GPIO config on all the tester helper devices (BBB's) in bad state
which caused other PRs failures.
Config is now restored so it should not cause failures now.

epyeoh:
@okartau Noted, thanks!

epyeoh:
test this please
mraa ssh related fail in daft testing, not related to this PR

PR #278, opened 2017-08-03
librealsense_%.bbappend: remove x11 checking

YinThong:
Require librealsense example and tutorial application be to install inside development image target.
In order to test librealsense on meta-iotqa.


okartau:
ok to test

ipuustin:
`librealsense` graphical examples have `glfw` as a dependency. If you look at `glfw` recipe you'll see that it in turn depends on `libxi`, `libxcursor`, `libxinerama`, `libxrandr` and so on. You shouldn't try enabling graphical examples without having `x11` distro feature enabled. I guess this builds now because there isn't a distro feature called `opengl-native` so graphical examples stay disabled.

I guess you are trying to get the graphical examples working on Wayland, because you didn't want to specify `x11` distro feature? You might want to look into how `glfw` could be made to work with Wayland.

YinThong:
Ok.. seem like i change in wrong direction. Let me concentrate on the multiconfig:x11:refkit-image-computervision-x11 creation in CI. Please kindly close this PR.

PR #282, opened 2017-08-07
Remove gtk dependency from opencv

juimonen:


ipuustin:
Previous additions to OpenCV `PACKAGECONFIG` are here: https://github.com/intel/intel-iot-refkit/blob/master/meta-refkit-core/conf/distro/include/refkit-config.inc#L147

juimonen:
so whats the logic of putting certain things there and certain things here (as bbappends)?

jlaako:
Having PACKAGECONFIG_remove_pn-opencv_df-refkit-config = "gtk" in refkit-config.inc works

juimonen:
ok, this new version should work, although not sure about the correct place of the fix...

mythi:
My preference is to keep all ```PACKAGECONFIG``` changes in one place (in our case refkit-config.in) for the sake of clarity and readability.

juimonen:
ok, moved to refkit-config.inc

mythi:
@juimonen the change looks good to me but the commit message is missing the *why* part. is it pulling something we don't want?

jlaako:
gtk-3 pulls in lot of dependencies, so we may want to be careful when/if we want to pull it in

PR #284, opened 2017-08-07
Test build: add new computer vision packages.

ipuustin:


meta-refkit-extra/recipes-convnet/darknet/files/0001-Makefile-enable-OpenMP-support.patch
```diff
@@ -0,0 +1,28 @@
...
+diff --git a/Makefile b/Makefile
+index b0a9f1c..17dd765 100644
+--- a/Makefile
++++ b/Makefile
+@@ -1,7 +1,7 @@
+ GPU=0
+ CUDNN=0
+ OPENCV=0
+-OPENMP=0
++OPENMP=1
```
mythi: Would this work without the patch by adding the new val in ```EXTRA_OEMAKE```

ipuustin: Yes, you're right.

ipuustin:
Let's add these piece by piece.

PR #291, opened 2017-08-16
refkit: move to meta-clang pyro branch

mythi:
meta-clang master moved to clang 5.0 release branch but
beignet fails to build with it so move to use meta-clang
pyro branch that has 4.0.


ipuustin:
Looks good to me. I hope upstream will keep updating pyro branch.

PR #298, opened 2017-08-22
supportedrecipes.py: fix reporting for virtual recipes

mythi:
With sources reporting enabled, check_build() fails to
find pn data for virtual target recipes (e.g., IMAGE_MODE
variant recipes).

Fix the failure by adding a check for 'virtual:' prefixed
fn's too. strip_multiconfig_prefix() becomes roughly equivalent
to bitbake's virtualfn2realfn().


okartau:
test this please, ci failure not related to change?

PR #299, opened 2017-08-22
CI : Fix attempt to egrep missing file, avoid publisher error on missing sdk/*.sh

okartau:
1st commit:
Recent addition about exposing traceback in main log
created problem where egrep was tried on missing file
in qemu tester case. Shell AND-ed operators are not
performed by left-to-right logic, commands are attempted first.

2nd commit avoids publisher error in case of missing sdk/*.sh
which can happen in certain build failure case.

### DIFF
--- a/docker/publish-project.sh
+++ b/docker/publish-project.sh
@@ -78,9 +78,9 @@ if [ -d ${_DEPL}/swupd/${TARGET_MACHINE} ]; then
 fi
 
 if [ -d ${_DEPL}/sdk ]; then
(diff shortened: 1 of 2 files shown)


okartau:
verfied in build log, no attempt

okartau:
2nd commit here causes BB server malfunctioning

okartau:
test this please

okartau:
adding test-build label, as this now contains a workaround (3rd commit), which likely is not available in upstream yet (but coming), so we should not merge as it is now

PR #301, opened 2017-09-01
Experimental: Use bb server timeout

okartau:
CI env: set BB_SERVER_TIMEOUT=5 for series of bitbake

NOTE: hopefully not needed, idea is that builds should work without it.
    
    During CI build there are now 26 bitbake invocations,
    and we see races involving old server exiting while a new BB
    trying to reconnect. Keeping server around for few seconds
    should help with those.


okartau:
test this please, unrelated error

okartau:
test this please, proxy error

okartau:
BB server race has not been spotted in logs after recent changes to related parts of code, thus closing this PR, we dont need this change now

PR #307, opened 2017-09-13
Try: build rootfs from rpms

mythi:


mythi:
now works as expected with the latest oe-core.

PR #324, opened 2017-11-03
meta-iotqa: upsquared.mask

samniem:
Ignores tests for UP Squared that cannot be run on it.

Jenkins uses mask files to determine the DAFT tests that are to be excluded for each supported board. UP Squared doesn't have Bluetooth and WIFI so it makes sense to exclude these tests to avoid unnecessary error notifications which would appear with every pull request test.

Signed-off-by: Sami Nieminen <sami1.nieminen@intel.com>

### DIFF
--- a/meta-iotqa/conf/test/upsquared.mask
+++ b/meta-iotqa/conf/test/upsquared.mask
@@ -1,0 +1,3 @@
+#this file contains tests which cant be run on Upsquared
+oeqa.runtime.sanity.comm_wifi_connect
+oeqa.runtime.sanity.comm_btcheck


samniem:
Test still needs to be run on upsquared device to ensure it is working as intended.

mythi:
@samniem please add the reasoning *why* they need to be masked in the commit message

samniem:
@mythi Added reasoning

mythi:
@samniem the commit is not updated

mythi:
@samniem to update the PR commit, you'd have to force push to your remote branch. There's no need to close the PR. If you're still planning to submit the change, you can reopen this PR and do the force push (the order is important)

samniem:
@mythi at the moment the pull request has to wait, as mraa and upm libraries need to be upgraded for UP Squared to work properly. UP Squared mask could be added to the same PR with these updates to make a neat package for UP Squared.

PR #326, opened 2017-11-10
support whole-disk encryption with TPM 2.0

pohly:
Besides enabling and testing TPM 2.0 support for whole-disk
encryption, this PR also enhances how refkit handles kernel
configuration.


pohly:
@mythi can you have a look at the PR?

It passed all tests now and includes a workaround for the unbuildable librealsense. The test_comm_btcheck looks unrelated.

test this please

meta-refkit-core/recipes-security/swtpm2/swtpm2_git.bb
```diff
@@ -0,0 +1,61 @@
+SUMMARY = "SWTPM2.0 - Software TPM2.0 Emulator"
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=fe8092c832b71ef20dfe4c6d3decb3a8"
+SECTION = "apps"
+
+DEPENDS = "libtasn1 fuse expect socat glib-2.0 libtpms2"
+
```
mythi: how about adding ```PROVIDES``` for these recipes? We'd avoid making changes to '2' and back.

meta-refkit/conf/distro/refkit.conf
```diff
@@ -187,6 +187,19 @@ BBMASK += " \
     meta-intel/common/recipes-core/ovmf \
 "
 
+# Without this we end up taking TPM2.0 recipes from meta-security
+# instead of meta-measured. We only need meta-tpm because of swtpm.
+# If we move that to meta-measured, we could instead drop meta-tpm.
+BBMASK += "meta-security/meta-tpm/recipes-tpm/tpm2"
```
mythi: With swtpm2 and others copied to meta-refkit, do we need meta-security at all? Dropping it could simplify ```BBMASK```ing.

meta-refkit-core/recipes-images/images/refkit-installer-image.bb
```diff
@@ -63,39 +63,89 @@ REFKIT_INSTALLER_UEFI_COMBO () {
...
+                    #
+                    # We store a random key in the TPM NVRAM where it is accessible
+                    # to the initramfs. The initramfs will turn off read-access
+                    # after it has retrieved the key, so nothing else that gets started
+                    # later will have access to the key.
+                    #
+                    # -t 0x80020002 = TPMA_NV_OWNERWRITE|TPMA_NV_OWNERREAD|TPMA_NV_READ_STCLEAR
+                    # (https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf)
+                    #
+                    # tpm2.0-tools master supports --attibutes=ownerwrite|ownerread|read_stdclear, but
```
mythi: Just for my own education: is master same as road to 3.x (referenced by the initramfs code)? Is that going to be available soon?

pohly: There is now a 3.0rc1 based on master, so 3.0 might not be too far away.

meta-refkit-core/recipes-security/swtpm2/swtpm2_git.bb
pohly: You mean, override swtpm*.bb with swtpm2*.bb? I didn't want to do that because I want to keep the choice open between using one or the other.

meta-refkit/conf/distro/refkit.conf
pohly: meta-security is still the place from where we get the TPM 1.2 recipes plus some other things, like keyutils.

mythi: OK, at quick glance I did not find any other use for meta-security than meta-tpm (supported recipes says ```keyutils@security-framework```)

meta-refkit-core/recipes-security/swtpm2/swtpm2_git.bb
mythi: Yes. My thinking was to make the selection easier for the end-user. Now modifications are needed in *many* places.

meta-refkit-core/recipes-images/images/refkit-installer-image.bb
mythi: Thanks!

meta-refkit-core/recipes-kernel/linux/refkit-kernel-cache/tpm2.cfg
```diff
@@ -0,0 +1,15 @@
...
+. an assembly of processed fragments, each of which is designed to
+. capture a specific part of the final configuration (e.g. platform
+. configuration, feature configuration, and board specific hardware
+. configuration).  For more information on kernel configuration, please
+. consult the product documentation.
+.
+..........................................................................
+
+CONFIG_TCG_TPM=y
+CONFIG_TCG_CRB=m
```
mythi: Looks like we need ```CONFIG_TCG_CRB=y``` or have the module installed and loaded for fTPM to get the device interface available in initramfs.

pohly:
On Mon, 2017-12-04 at 08:52 +0000, Mikko Ylinen wrote:
> +CONFIG_TCG_TPM=y
> +CONFIG_TCG_CRB=m
> 
> Looks like we need CONFIG_TCG_CRB=y or have the module installed and
> loaded for fTPM to get the device interface available in initramfs.

Building it into the kernel is less intrusive IMHO. Adding the right
set of kernel modules into the initramfs would be harder.

mythi:
On Mon, Dec 4, 2017 at 11:18 AM, Patrick Ohly wrote:

> On Mon, 2017-12-04 at 08:52 +0000, Mikko Ylinen wrote:
> > +CONFIG_TCG_TPM=y
> > +CONFIG_TCG_CRB=m
> >
> > Looks like we need CONFIG_TCG_CRB=y or have the module installed and
> > loaded for fTPM to get the device interface available in initramfs.
>
> Building it into the kernel is less intrusive IMHO. Adding the right
> set of kernel modules into the initramfs would be harder.
>
>

I thought so too. Can you make the change?

pohly:
On Mon, 2017-12-04 at 11:54 +0000, Mikko Ylinen wrote:
> > Building it into the kernel is less intrusive IMHO. Adding the
> right
> > set of kernel modules into the initramfs would be harder.
> >
> >
> 
> I thought so too. Can you make the change?

Done.

In the meantime, Stefan Berger has also added tags to his latest tpm2
preview branches, so I was able to switch back to just using the
upstream repos in the recipes.

Perhaps this can be merged now?

mythi:
@pohly Great, thanks! Looks good to me. There is one more thing to check before merging. I noticed this when there was a failure:

an empty ```$keyfile``` flows through until ```cryptsetup open``` (and ```luks_cleanup```) if neither ```/dev/tpm*``` is found nor ```REFKIT_DISK_ENCRYPTION_PASSWORD``` is set.  It triggers faulty parameters to dd at least: ```bs="$(stat -c '%s' "$keyfile")``` (bs=0).

pohly:
On Mon, 2017-12-04 at 23:57 -0800, Mikko Ylinen wrote:
> I noticed this when there was a failure:
> an empty $keyfile flows through until cryptsetup open (and
> luks_cleanup) if neither /dev/tpm* is found nor
> REFKIT_DISK_ENCRYPTION_PASSWORD is set.

I'll add a warning message about the empty keyfile. It's not necessary
a hard error (but close), so calling "fatal" might be a bit too
drastic.

>  It triggers faulty parameters to dd at least: bs="$(stat -c '%s'
> "$keyfile") (bs=0).

I'll add an if check for that.

PR #329, opened 2017-11-21
meta-flatpak: fix ostree build failure with updated glib-2.0 (2.54.2).

klihub:
Updated ostree to the latest tagged upstream release v2017.13.
This should fix libglnx compilation problems when compiling
against glib-2.54.2 or newer.

The recently added bash completion has been split out to a
submodule of its own. If you have bash in your image and want
autocompletion for ostree, add the submodule to the image.

### DIFF
--- a/meta-flatpak/recipes-ostree/ostree/ostree_git.bb
+++ b/meta-flatpak/recipes-ostree/ostree/ostree_git.bb
@@ -9,9 +9,9 @@ SRC_URI = " \
     file://0001-autogen.sh-fall-back-to-no-gtkdocize-if-it-is-there-.patch \
 "
 
-SRCREV = "5a5e465492aca13937dab7a2df39f25da94e6e36"
+SRCREV = "ae61321046ad7f4148a5884c8c6c8b2594ff840e"
 
-PV = "2017.8+git${SRCPV}"
+PV = "2017.13+git${SRCPV}"
 S = "${WORKDIR}/git"
 
 inherit autotools pkgconfig requires-systemd gobject-introspection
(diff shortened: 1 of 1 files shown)


pohly:
Why are  the tmpfiles not needed?

klihub:
AFAICT, it is a tmpfiles config file to clean up lingering ostree overlay(fs) lock files. We don't use ostree with overlays, hence we don't need to clean up lingering lock files. But I guess we could leave that config in place as well.

pohly:
Please add a comment explaining why the files are split out and when they should be installed.

klihub:
I updated the patch to include the ostree tmpfiles.d config fragment in the base package. It's safe to have there (IOW it should not conflict/interfere with our slightly different way of booting into an ostree-enabled image), it's a small one-liner, and this way there is less logic involved in deciding which ostree packages you need in an image.

meta-flatpak/recipes-ostree/ostree/ostree_git.bb
```diff
@@ -9,7 +9,7 @@ SRC_URI = " \
     file://0001-autogen.sh-fall-back-to-no-gtkdocize-if-it-is-there-.patch \
 "
 
-SRCREV = "5a5e465492aca13937dab7a2df39f25da94e6e36"
+SRCREV = "ae61321046ad7f4148a5884c8c6c8b2594ff840e"
 
 PV = "2017.8+git${SRCPV}"
```
mythi: PV should also be updated to match v2017.13 tag.

klihub:
Updated PV to match version tag.

PR #318, opened 2017-10-05
TEST: swupd support and testing

pohly:
We are investigating supporting swupd again. This is a test PR using
an experimental fork of meta-swupd.

PR #330, opened 2017-11-27
meta-intel-realsense update

pohly:
This change is necessary to avoid build errors in PR #318 and PR #326
and probably everything else that triggers a rebuild of librealsense.


mythi:
@pohly meta-intel-librealsense update did not help?

pohly:
On Mon, 2017-11-27 at 10:21 +0000, Mikko Ylinen wrote:
> @pohly meta-intel-librealsense update did not help?

I thought we were stuck at the current version? Or is that some other
sub-component (meta-clang)?

Looking at the latest master, it looks like it did change the branch,
so updating should work. I'll use this PR to try that.

ipuustin:
The `legacy` branch. :-)

mythi:
@pohly thanks! I believe it's meta-clang you were thinking

@ipuustin looks like they've added librealsense2

PR #331, opened 2017-11-28
Remove ovmf secure boot patch

jlaako:
ovmf UEFI secure boot patch is now provided by upstream meta-intel,
not needed in refkit anymore.


mythi:
The meta-intel ovmf bbappend is in BBMASK. The mask setting probably fails with the latest meta-intel that dropped /common/ directory. Perhaps just update BBMASK and pull in the meta-intel update?

I'd think it's better to use the patch in this repo because this repo also ships the 'development' key the patch uses. meta-intel does not have that development key anywhere AFAICS.

jlaako:
@mythi OK! Updated the PR accordingly...

PR #336, opened 2018-04-01
openblas: Upgrade to 0.2.20 and fix cross builds

kraj:
Its under BSD-3-Clause to make LICENSE reflect that
Define PREFIX and CROSS_PREFIX to get the tooling right

Signed-off-by: Khem Raj <raj.khem@gmail.com>


ipuustin:
Thanks! This appears to build fine and the changes (including cleanups) make sense to me. LGTM.

mythi (approved):

mythi:
@kraj thanks. have you noted meta-convnet? that might be better suited for your needs (although it seems a bit outdated too).

